PR #76, opened 2022-03-11
Make default qos rate unit bps and setUESubMbr

thakurajayL:
QoS rate default unit bps
QoS UESubscribedAMBR is set to single device group qos.


thakurajayL:
retest this please

PR #80, opened 2022-10-10
keep connecting until it connects to MongoDB then listen on 5000 port for configuration from Roc/Simapp

vthiruveedula:
1. keep connecting until it connects to MongoDB
2. go version to 1.16


thakurajayL:
+1

thakurajayL (approved):

PR #84, opened 2023-11-10
Adds subscribers to Am data collection when they are created.

saltiyazan:
To be able to get subscribers created using `PostSubscriberByID` and making the behavior of `GetSubscribers` symmetric to it, we are adding a subscriber entry to the `subscriptionData.provisionedData.amData` collection. At the time of the creation of the subscriber only the `imsi` is known and is added in the `ueId` field, the rest of the fields are updated later on.


saltiyazan:
Please take a look, if it seems good I will put it to ready to review @ghislainbourgeois :)

configapi/api_sub_config.go
```diff
@@ -275,15 +275,18 @@ func GetSubscribers(c *gin.Context) {
...
 	for _, amData := range amDataList {
-		ueId := amData["ueId"]
-		servingPlmnId := amData["servingPlmnId"]
-		tmp := configmodels.SubsListIE{
-			PlmnID: servingPlmnId.(string),
-			UeId:   ueId.(string),
+		ueId := amData["ueId"].(string) // Since ueId is guaranteed to exist
+
+		var tmp configmodels.SubsListIE
+		tmp.UeId = ueId
```
ghislainbourgeois, line 284:
```suggestion
		tmp := configmodels.SubsListIE{
		    UeId: amData["ueID"].(string),
		}
```

ghislainbourgeois (changes requested):
Looks good, I only have one small comment

configapi/api_sub_config.go
saltiyazan: ✅

ghislainbourgeois (approved):

configapi/api_sub_config.go
```diff
@@ -275,15 +275,18 @@ func GetSubscribers(c *gin.Context) {
...
+	var subsList []configmodels.SubsListIE
 	amDataList := MongoDBLibrary.RestfulAPIGetMany(amDataColl, bson.M{})
 	for _, amData := range amDataList {
-		ueId := amData["ueId"]
-		servingPlmnId := amData["servingPlmnId"]
+
 		tmp := configmodels.SubsListIE{
-			PlmnID: servingPlmnId.(string),
-			UeId:   ueId.(string),
+		    UeId: amData["ueID"].(string),
```
gruyaume, line 283:
```suggestion
		    UeId: amData["ueId"].(string),
```

gruyaume (changes requested):

gab-arrobo:
@saltiyazan please rebase this PR

saltiyazan:
> > > @saltiyazan please rebase this PR
> > 
> > 
> > Just did. @gab-arrobo :)
> 
> Hi @saltiyazan, it is not. You need to align it with the master branch and that is why GitHub is showing `This branch is out-of-date with the base branch`
> 
> image

Hi @gab-arrobo 
How about now? I think it's in sync now.

gab-arrobo:
test this please

proto/server/configEvtHandler_test.go
```diff
@@ -272,3 +272,30 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
 		}
 	}
 }
+
+func Test_handleSubscriberPost(t *testing.T) {
+	imsi := "1234"
```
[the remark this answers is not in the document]
saltiyazan: If I'm not mistaken the `ueId` is provided without the `imsi-` prefix in the post request, example:
```
curl -v ${WEBUI_IP}:5000/api/subscriber/imsi-208930100007487 \
--header 'Content-Type: text/plain' \
--data '{
    "UeId":"208930100007487",
    "opc":"981d464c7c52eb6e5036234984ad0bcf",
    "key":"5122250214c33e723a5dd523fc145fc0",
    "sequenceNumber":"16f3b3f70fc2"
}
```
`PostSubscriberByID` then takes the `ueId` and passes it in the `configChannel` channel as the imsi.
And then the `imsi-` prefix is added in the functions of `configEvtHandler.go` as in `configHandler`
Based on that I think the imsi here should given and put into `configMsg` without the prefix.
The name `imsi` might be confusing, I changed it to `ueId` and changed `1234` to `208930100007487`
Please let me know if I'm missing something or being wrong.

gab-arrobo:
ok to test

proto/server/configEvtHandler.go
```diff
@@ -163,6 +164,17 @@ func configHandler(configMsgChan chan *configmodels.ConfigMessage, configReceive
 	}
 }
 
+func handleSubscriberPost(configMsg *configmodels.ConfigMessage) {
```
gab-arrobo, line 167: Should we also add the `get` method(s) for this? The other `handle` functions have `get` functions.

saltiyazan: That is because the "subscriber" is handled a bit differently in the code.
The subscriber object does not exist as such, when a call is made to `get subscriber by ID`, the object is created by fetching data from different collections, and put into the `SubsData` object.
When we make a `post call to create the subscriber`, we don't really create the subscriber, we post a message with the `ueId` to `configChannel` and then update the `AuthenticationSubscription` data collection. Only when the device group and network slice are created the rest of the data required to build the `SubsData` object is added.
In this PR my aim is to save the `ueId` to the `AmData` even before creating the device group and network slice so we do get a list of the subscribers when we call `GetSubscribers`.
Having that said, based on my understanding of how subscribers are handled in the code, adding a `getSubscriber` here would be misleading, as we can't be sure of having all ghe bits needed ti build the `SubsData` right after a `subscriber post`, all we have for sure is the `ueId`.
The handler of the `get request for a subscriber` is in `api_sub_config.go` file as `GetSubscriberByID` and `GetSubscribers`.

Please let me know if I have answered the question or not, or if I'm missing anything here :)

thakurajayL (approved):

thakurajayL:
Could you please rebase your branch,

PR #90, opened 2023-12-01
chore: Adds dependabot.yaml config file

gruyaume:
## Description

This PR adds a dependabot.yaml file so that Dependabot can come and propose library updates to the project through Pull Requests. This prevents having developers regularly looking up for library changes and avoids distributing code that should have been patched. Dependabot is also used in other SD-Core projects like the UPF and the config file here was taken from there.

For more information about dependabot:
- https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates


gab-arrobo (approved):
+1

PR #104, opened 2023-12-05
chore: Adds build job to workflow

gruyaume:
## Description

This PR adds a build job in addition to the existing unit tests run. It also uniquely relies on `go test` instead of `make test` which would deploy a docker container and run `go test` in it.


gab-arrobo (approved):
+1 (Tested changes locally)

gab-arrobo:
ok to test

PR #113, opened 2024-01-05
chore: Uses go version from go.mod

gruyaume:
# Description

Uses the go version from the go.mod file instead of freezing it to 1.21 in the CI.


gab-arrobo:
ok to test

gab-arrobo (approved):
+1

PR #119, opened 2024-01-10
chore: Adds govet lint check in CI

patriciareinoso:
Adds the go vet lint check to the CI
Enables all analyzers 
`fieldalignment`  defines an Analyzer that detects structs that would use less memory if their fields were sorted.


gab-arrobo:
ok to test

proto/server/clientEvtHandler.go
```diff
@@ -94,29 +94,29 @@ type userPlaneProfile struct {
...
-	SubSelectRules    []*subSelectionRule          `json:"subscriber-selection-rules,omitempty"`
 	ApnProfiles       map[string]*apnProfile       `json:"apn-profiles,omitempty"`
 	UserPlaneProfiles map[string]*userPlaneProfile `json:"user-plane-profiles,omitempty"`
 	QosProfiles       map[string]*qosProfile       `json:"qos-profiles,omitempty"`
+	SubSelectRules    []*subSelectionRule          `json:"subscriber-selection-rules,omitempty"`
 }
 
 type configHss struct {
+	Key         string                 `json:"Key,omitempty"`
+	ApnProfiles map[string]*apnProfile `json:"apn-profiles,omitempty"`
```
gab-arrobo, line 111: Why not the following?
```suggestion
	ApnProfiles map[string]*apnProfile `json:"apn-profiles,omitempty"`
	Key         string                 `json:"Key,omitempty"`
```

configmodels/config_msg.go
```diff
@@ -22,14 +22,14 @@ const (
 )
 
 type ConfigMessage struct {
-	MsgType      int
-	MsgMethod    int
 	DevGroup     *DeviceGroups
 	Slice        *Slice
 	DevGroupName string
 	SliceName    string
 	AuthSubData  *models.AuthenticationSubscription
```
gab-arrobo, line 29: Should not this also be changed?
```suggestion
	AuthSubData  *models.AuthenticationSubscription
	DevGroupName string
	SliceName    string
```

gab-arrobo (commented):
Have you run a test with these changes to make sure that nothing breaks due to the order of the fields?

gab-arrobo:
@patriciareinoso, any update about this PR?

configmodels/config_msg.go
patriciareinoso: I don't get any error when running `golangci-lint run` 
Also from what I have seen pointers and string take the same place so there would not be any change on the total size of the struct 
But I agree it would look cleaner

proto/server/clientEvtHandler.go
patriciareinoso: Same comment as above

PR #242, opened 2024-10-25
Inventory clean-up

patriciareinoso:
This PR:
- Adds documentation to inventory API handlers
- Simplifies the inventory POST handlers
- Adds a new type to handle POST UPF and POST gNB request
- Refactor UTs to make them table-driven
- Replace tabs on godoc comments by spaces to properly align fields
- Add a job to check that building the project using the `ui` tag is not broken

### DIFF
--- a/configapi/api_inventory.go
+++ b/configapi/api_inventory.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 
 	"github.com/gin-gonic/gin"
-	"github.com/omec-project/util/httpwrapper"
 	"github.com/omec-project/webconsole/backend/logger"
 	"github.com/omec-project/webconsole/configmodels"
 	"github.com/omec-project/webconsole/dbadapter"
@@ -30,16 +29,28 @@ func setInventoryCorsHeader(c *gin.Context) {
 	c.Writer.Header().Set("Access-Control-Allow-Methods", "POST, GET, DELETE")
 }
 
+// GetGnbs godoc
+//
+//	@Description	Return the list of gNBs
+//	@Tags			gNBs
+//	@Produce		json
+//	@Security		BearerAuth
+//	@Success		200	{array}		configmodels.Gnb	"List of gNBs"
+//	@Failure		401	{object}	nil					"Authorization failed"
+//	@Failure		403	{object}	nil					"Forbidden"
+//	@Failure		500	{object}	nil					"Error retrieving gNBs"
+//	@Router			/config/v1/inventory/gnb	[get]
 func GetGnbs(c *gin.Context) {
 	setInventoryCorsHeader(c)
-	logger.WebUILog.Infoln("Get all gNBs")
+	logger.WebUILog.Infoln("get all gNBs")
 
 	var gnbs []*configmodels.Gnb
 	gnbs = make([]*configmodels.Gnb, 0)
 	rawGnbs, errGetMany := dbadapter.CommonDBClient.RestfulAPIGetMany(gnbDataColl, bson.M{})
 	if errGetMany != nil {
 		logger.DbLog.Errorln(errGetMany)
-		c.JSON(http.StatusInternalServerError, gnbs)
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to retrieve gNBs"})
+		return
 	}
 
 	for _, rawGnb := range rawGnbs {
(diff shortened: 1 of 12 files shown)


configapi/api_inventory.go
```diff
@@ -30,16 +29,28 @@ func setInventoryCorsHeader(c *gin.Context) {
...
+//
+//	@Description	Return the list of gNBs
+//	@Tags			gNBs
+//	@Produce		json
+//	@Security		BearerAuth
+//	@Success		200	{array}		configmodels.Gnb	"List of gNBs"
+//	@Failure		401	{object}	nil					"Authorization failed"
+//	@Failure		403	{object}	nil					"Forbidden"
+//	@Failure		500	{object}	nil					"Error retrieving gNBs"
+//	@Router			/config/v1/inventory/gnb	[get]
```
gab-arrobo, line 42: Do you need to use "tabs" here? What about using whitespaces instead of tabs to properly align the text?
Note: This comment is applicable to all doc comments
```suggestion
// @Description	Return the list of gNBs
// @Tags      gNBs
// @Produce   json
// @Security  BearerAuth
// @Success   200     {array}   configmodels.Gnb  "List of gNBs"
// @Failure   401     {object}  nil               "Authorization failed"
// @Failure   403     {object}  nil               "Forbidden"
// @Failure   500     {object}  nil               "Error retrieving gNBs"
// @Router    /config/v1/inventory/gnb  [get]
```

configapi/api_inventory.go
```diff
@@ -72,49 +108,44 @@ func DeleteGnb(c *gin.Context) {
...
 
 func handlePostGnb(c *gin.Context) error {
-	var gnbName string
-	var exists bool
-	if gnbName, exists = c.Params.Get("gnb-name"); !exists {
+	gnbName, exists := c.Params.Get("gnb-name")
+	if !exists {
 		errorMessage := "post gNB request is missing gnb-name"
 		logger.ConfigLog.Errorln(errorMessage)
 		return errors.New(errorMessage)
```
gab-arrobo, line 115: What about using a single way to "report" errors? I see you are using `fmt.Errorf` but some previous code is using `errors.New` in other places
```suggestion
		return fmt.Errorf(errorMessage)
```

configapi/api_inventory.go
```diff
@@ -62,6 +85,18 @@ func PostGnb(c *gin.Context) {
...
+//	@Description  Delete an existing gNB
+//	@Tags         gNBs
+//	@Produce      json
+//	@Param        gnb-name    path    string    true    "Name of the gNB"
+//	@Security     BearerAuth
+//	@Success      200  {object}  nil  "gNB deleted"
+//	@Failure      400  {object}  nil  "Failed to delete the gNB"
+//	@Failure      401  {object}  nil  "Authorization failed"
+//	@Failure      403  {object}  nil  "Forbidden"
+//	@Router       /config/v1/inventory/gnb/{gnb-name}  [delete]
```
gab-arrobo, line 99:
```suggestion
// @Description  Delete an existing gNB
// @Tags         gNBs
// @Produce      json
// @Param        gnb-name    path    string    true    "Name of the gNB"
// @Security     BearerAuth
// @Success      200  {object}  nil  "gNB deleted"
// @Failure      400  {object}  nil  "Failed to delete the gNB"
// @Failure      401  {object}  nil  "Authorization failed"
// @Failure      403  {object}  nil  "Forbidden"
// @Router       /config/v1/inventory/gnb/{gnb-name}  [delete]
```

configapi/api_inventory.go
```diff
@@ -72,52 +107,47 @@ func DeleteGnb(c *gin.Context) {
...
 		errorMessage := "post gNB request body is missing tac"
 		logger.ConfigLog.Errorln(errorMessage)
-		return errors.New(errorMessage)
+		return fmt.Errorf("%s", errorMessage)
+	}
+	postGnb := configmodels.Gnb{
+		Name: gnbName,
+		Tac:  postGnbRequest.Tac,
 	}
-	req := httpwrapper.NewRequest(c.Request, newGnb)
```
gab-arrobo, line 100: I see you are removing the `httpwrapper` (github.com/omec-project/util/httpwrapper) from this file. Does it mean that it is "obsolete" and the way you are doing it now (`postGnb`) is better? If so, is this something that should be done across the whole 5GC codebase? For example, the `httpswrapper` is also used in `configapi/api_slice_mgmt.go`, should we follow the same approach there as is done here (configapi/api_inventory.go)?

gab-arrobo (approved):
+1

configapi/api_inventory.go
patriciareinoso: That depends on what you need to do with the new request created by `httpwrapper.NewRequest`. On this case, getting the Gnb using `c.ShouldBindJSON`, then creating a new request using the `httpwrapper` just to extract the Gnb again, seemed unnecessary. Not using the `httpwrapper` is more straightforward. 
If you use the `httpwrapper` to somehow filter the content on the body and use that request for something else again, sure, let's keep it. But if we just need to extract the important content from the body, I think we can remove it.

PR #244, opened 2024-11-02
Refactor path for configuration file, option(s) and Docker image

gab-arrobo:


thakurajayL:
As such looks good.. Will this break any other work for Canonical?

gab-arrobo:
> As such looks good.. Will this break any other work for Canonical?

I do not think this should break anything. However, I can double check with them. The changes here are consistent with the other NFs. So, if these changes break something on the Canonical side, the changes in all of the other NFs will have similar effect

configmodels/model_slice.go
```diff
@@ -15,7 +15,7 @@
 package configmodels
 
 type Slice struct {
-	SliceName string `json:slice-name,omitempty"`
+	SliceName string `json:"slice-name,omitempty"`
```
gab-arrobo, line 18: @thakurajayL, 
FYI, the tests fail because of this change

gab-arrobo: This is fixed now

gab-arrobo:
@patriciareinoso, any objection about this PR?

patriciareinoso:
We are currently using `/bin/webconsole --webuicfg <path>` to run the service (same for the other NFs),  so it will break our stuff. However it's only a small change to take care of when updating the images.

thakurajayL:
@patriciareinoso its not just webconsole but also other NFs have gone through similar change.

thakurajayL (approved):

PR #250, opened 2024-11-07
feat: Adds Pebble notifications

Gmerold:
This PR adds Pebble notifications whenever there's a change against the NetworkSlice.
Additionally, a minor refactoring has been done to prevent running logic for removing NetworkSlices while removing DeviceGroups.

CONTEXT
[Pebble](https://github.com/canonical/pebble) is a service manager widely used in Canonical's solutions (i.e. SD-Core charms). 
We're proposing this change to be able to orchestrate network configuration between the Core and the RAN. [Pebble notices](https://canonical-pebble.readthedocs-hosted.com/en/latest/reference/cli-commands/notify/#reference-pebble-notify-command) send a custom event which can be observed by the [charmed operators](https://juju.is/docs/juju/charmed-operator). Based on that we'll be able to run some logic in the charmed operator which will configure the RAN part of the network.

Pebble notices feature is disabled by default. To enable it, additional configuration parameter needs to be set. This way, the impact on other parts of the `webconsole` are minimal.


proto/server/configEvtHandler.go
```diff
@@ -210,6 +215,28 @@ func handleNetworkSlicePost(configMsg *configmodels.ConfigMessage, subsUpdateCha
 	if errPost != nil {
 		logger.DbLog.Warnln(errPost)
 	}
+	if factory.WebUIConfig.Configuration.SendPebbleNotifications {
+		sendPebbleNotification("canonical.com/webconsole/networkslice/create")
```
gab-arrobo, line 219: Should not this be a configurable parameter instead of a hard-coded value?

proto/server/configEvtHandler.go
```diff
@@ -210,6 +215,28 @@ func handleNetworkSlicePost(configMsg *configmodels.ConfigMessage, subsUpdateCha
...
+		config5gMsg.PrevSlice = getSliceByName(configMsg.SliceName)
+		subsUpdateChan <- &config5gMsg
+	}
+	filter := bson.M{"SliceName": configMsg.SliceName}
+	errDelOne := dbadapter.CommonDBClient.RestfulAPIDeleteOne(sliceDataColl, filter)
+	if errDelOne != nil {
+		logger.DbLog.Warnln(errDelOne)
+	}
+	if factory.WebUIConfig.Configuration.SendPebbleNotifications {
+		sendPebbleNotification("canonical.com/webconsole/networkslice/delete")
```
gab-arrobo, line 238: Should not this be a configurable parameter instead of a hard-coded value?

proto/server/configEvtHandler.go
Gmerold: At this point I don't see a need for parameterizing the key, because our charmed operators will be the only user of this feature. In the future, if it becomes troublesome, we could think about building this string out of configurable elements.

proto/server/configEvtHandler.go
Gmerold: At this point I don't see a need for parameterizing the key, because our charmed operators will be the only user of this feature. In the future, if it becomes troublesome, we could think about building this string out of configurable elements.

proto/server/configEvtHandler.go
```diff
@@ -210,6 +215,34 @@ func handleNetworkSlicePost(configMsg *configmodels.ConfigMessage, subsUpdateCha
 	if errPost != nil {
 		logger.DbLog.Warnln(errPost)
 	}
+	if factory.WebUIConfig.Configuration.SendPebbleNotifications {
+		err := sendPebbleNotification("canonical.com/webconsole/networkslice/create")
```
thakurajayL, line 219: Along with Notification configuration, take the target base url configuration as well. So hardcoding specific to company name can be avoided.

proto/server/configEvtHandler.go
```diff
@@ -210,6 +215,34 @@ func handleNetworkSlicePost(configMsg *configmodels.ConfigMessage, subsUpdateCha
...
+	}
+	filter := bson.M{"SliceName": configMsg.SliceName}
+	errDelOne := dbadapter.CommonDBClient.RestfulAPIDeleteOne(sliceDataColl, filter)
+	if errDelOne != nil {
+		logger.DbLog.Warnln(errDelOne)
+	}
+	if factory.WebUIConfig.Configuration.SendPebbleNotifications {
+		err := sendPebbleNotification("canonical.com/webconsole/networkslice/delete")
+		if err != nil {
+			logger.ConfigLog.Warnln("sedning Pebble notification failed: %s. continuing silently.", err.Error())
```
thakurajayL, line 243:
```suggestion
			logger.ConfigLog.Warnln("sending Pebble notification failed: %s. continuing silently.", err.Error())
```

proto/server/configEvtHandler.go
```diff
@@ -196,6 +185,22 @@ func handleDeviceGroupPost(configMsg *configmodels.ConfigMessage, subsUpdateChan
 	rwLock.Unlock()
 }
 
+func handleDeviceGroupDelete(configMsg *configmodels.ConfigMessage, subsUpdateChan chan *Update5GSubscriberMsg) {
+	rwLock.Lock()
+	if factory.WebUIConfig.Configuration.Mode5G {
+		var config5gMsg Update5GSubscriberMsg
+		config5gMsg.Msg = configMsg
+		config5gMsg.PrevDevGroup = getDeviceGroupByName(configMsg.DevGroupName)
+		subsUpdateChan <- &config5gMsg
```
thakurajayL, line 194: I wonder was this code missing from old code? We were not writing event on channel?

Gmerold: It was always there. I just extracted the code responsible for deleting network slices and device groups to separate methods. IMO it reads better and makes it consistent with the rest of the code.

proto/server/configEvtHandler.go
Gmerold: This string is here just for namespacing the notifications. I really don't think it's worth adding another config param for it. This param will also need some validation, handling for the case when it's missing. 
I just committed a proposal to use `aetherproject.org` as the base URL to avoid "promoting" external companies ;) Would that work?

proto/server/configEvtHandler.go
ghislainbourgeois: I have the same question as Ajay, if the code was simply refactored, there should be an equivalent line removed elsewhere no?

ghislainbourgeois: Line 141-144 in your branch are writing to the channel, so with your change as it is right now, each change would be sent to the channel twice.

proto/server/configEvtHandler.go
```diff
@@ -210,6 +215,34 @@ func handleNetworkSlicePost(configMsg *configmodels.ConfigMessage, subsUpdateCha
...
+	if factory.WebUIConfig.Configuration.Mode5G {
+		var config5gMsg Update5GSubscriberMsg
+		config5gMsg.Msg = configMsg
+		config5gMsg.PrevSlice = getSliceByName(configMsg.SliceName)
+		subsUpdateChan <- &config5gMsg
+	}
+	filter := bson.M{"SliceName": configMsg.SliceName}
+	errDelOne := dbadapter.CommonDBClient.RestfulAPIDeleteOne(sliceDataColl, filter)
+	if errDelOne != nil {
+		logger.DbLog.Warnln(errDelOne)
```
ghislainbourgeois, line 238: With the refactor, no need for a different variable name for err:

```suggestion
	err := dbadapter.CommonDBClient.RestfulAPIDeleteOne(sliceDataColl, filter)
	if err != nil {
		logger.DbLog.Warnln(err)
```

proto/server/configEvtHandler.go
```diff
@@ -196,6 +185,22 @@ func handleDeviceGroupPost(configMsg *configmodels.ConfigMessage, subsUpdateChan
...
+	if factory.WebUIConfig.Configuration.Mode5G {
+		var config5gMsg Update5GSubscriberMsg
+		config5gMsg.Msg = configMsg
+		config5gMsg.PrevDevGroup = getDeviceGroupByName(configMsg.DevGroupName)
+		subsUpdateChan <- &config5gMsg
+	}
+	filter := bson.M{"group-name": configMsg.DevGroupName}
+	errDelOne := dbadapter.CommonDBClient.RestfulAPIDeleteOne(devGroupDataColl, filter)
+	if errDelOne != nil {
+		logger.DbLog.Warnln(errDelOne)
```
ghislainbourgeois, line 199: With the refactor, no need for a different variable name for `err`:

```suggestion
	err := dbadapter.CommonDBClient.RestfulAPIDeleteOne(devGroupDataColl, filter)
	if err != nil {
		logger.DbLog.Warnln(err)
```

ghislainbourgeois (changes requested):

proto/server/configEvtHandler.go
Gmerold: Ahh, of course. Copy-pasting backfired on me. 
Sorry for the confusion. I removed the entire `if` from both new functions, so that writing to the channel is handled centrally for all deletions.

ghislainbourgeois (approved):

gab-arrobo:
@Gmerold, please rebase your PR (resolve conflicts). Thanks!

Gmerold:
> @Gmerold, please rebase your PR (resolve conflicts). Thanks!

Hi @gab-arrobo, done .

backend/factory/config.go
```diff
@@ -31,14 +31,14 @@ type Info struct {
 }
 
 type Configuration struct {
-	Mongodb              *Mongodb    `yaml:"mongodb"`
-	RocEnd               *RocEndpt   `yaml:"managedByConfigPod,omitempty"` // fetch config during bootup
-	LteEnd               []*LteEndpt `yaml:"endpoints,omitempty"`          // LTE endpoints are configured and not auto-detected
-	TLS                  *TLS        `yaml:"tls"`
```
gab-arrobo, line 37: I think you unintentionally deleted this line

Gmerold: Indeed. Good catch. Fixed.

gab-arrobo (commented):
Overall, it looks go to me

proto/server/configEvtHandler.go
```diff
@@ -720,3 +741,14 @@ func SnssaiModelsToHex(snssai models.Snssai) string {
 	sst := fmt.Sprintf("%02x", snssai.Sst)
 	return sst + snssai.Sd
 }
+
+var execCommand = exec.Command
+
+func sendPebbleNotification(key string) error {
+	cmd := execCommand("pebble", "notify", key)
```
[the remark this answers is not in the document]
Gmerold: Without creating this variable I'm not able to mock the exec command. It's really only for testing purposes, but I believe it's a fair trade-off. WDYT?

backend/factory/config.go
```diff
@@ -31,14 +31,15 @@ type Info struct {
...
-	CfgPort              int         `yaml:"cfgport,omitempty"`
+	Mongodb                 *Mongodb    `yaml:"mongodb"`
+	RocEnd                  *RocEndpt   `yaml:"managedByConfigPod,omitempty"` // fetch config during bootup
+	LteEnd                  []*LteEndpt `yaml:"endpoints,omitempty"`          // LTE endpoints are configured and not auto-detected
+	TLS                     *TLS        `yaml:"tls"`
+	Mode5G                  bool        `yaml:"mode5G,omitempty"`
+	SdfComp                 bool        `yaml:"spec-compliant-sdf"`
+	EnableAuthentication    bool        `yaml:"enableAuthentication,omitempty"`
+	CfgPort                 int         `yaml:"cfgport,omitempty"`
+	SendPebbleNotifications bool        `yaml:"send_pebble_notifications,omitempty"`
```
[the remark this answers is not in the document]
Gmerold: And I definitely don't wanna mess with @patriciareinoso code ;)
Fixed.

proto/server/configEvtHandler.go
```diff
@@ -196,6 +185,16 @@ func handleDeviceGroupPost(configMsg *configmodels.ConfigMessage, subsUpdateChan
...
 
+func handleDeviceGroupDelete(configMsg *configmodels.ConfigMessage) {
+	rwLock.Lock()
+	filter := bson.M{"group-name": configMsg.DevGroupName}
+	err := dbadapter.CommonDBClient.RestfulAPIDeleteOne(devGroupDataColl, filter)
+	if err != nil {
+		logger.DbLog.Warnln(err)
+	}
+	rwLock.Unlock()
+}
```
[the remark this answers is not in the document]
Gmerold: Sorry bout that. I initially branched off using different base. Then, while rebasing I copy-pasted some code and that's how it backfires on me. 
I cleaned up the code (hopefully) and restored everything that's needed.

proto/server/configEvtHandler.go
```diff
@@ -210,6 +209,28 @@ func handleNetworkSlicePost(configMsg *configmodels.ConfigMessage, subsUpdateCha
...
+		logger.DbLog.Warnln(err)
+	}
+	if factory.WebUIConfig.Configuration.SendPebbleNotifications {
+		err := sendPebbleNotification("aetherproject.org/webconsole/networkslice/delete")
+		if err != nil {
+			logger.ConfigLog.Warnln("sending Pebble notification failed: %s. continuing silently.", err.Error())
+		}
+	}
 	rwLock.Unlock()
 }
```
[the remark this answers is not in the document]
Gmerold: Same reason as above. Sorry for the confusion.

proto/server/configEvtHandler.go
```diff
@@ -196,6 +182,22 @@ func handleDeviceGroupPost(configMsg *configmodels.ConfigMessage, subsUpdateChan
...
 
+func handleDeviceGroupDelete(configMsg *configmodels.ConfigMessage, subsUpdateChan chan *Update5GSubscriberMsg) {
+	rwLock.Lock()
+	if factory.WebUIConfig.Configuration.Mode5G {
+		var config5gMsg Update5GSubscriberMsg
+		config5gMsg.Msg = configMsg
+		config5gMsg.PrevDevGroup = getDeviceGroupByName(configMsg.DevGroupName)
+		subsUpdateChan <- &config5gMsg
+	}
```
gab-arrobo, line 192: In the previous code, the deletion operation (device group or slice) takes place first and then, the writing event on channel happens. Now, here, in the new functions, you first write event to channel and then do the delete operation(s). Should we keep the same order as before (delete operation(s) and then write event to channel)? or is the order not relevant?

proto/server/configEvtHandler.go
```diff
@@ -210,6 +212,34 @@ func handleNetworkSlicePost(configMsg *configmodels.ConfigMessage, subsUpdateCha
 	if errPost != nil {
 		logger.DbLog.Warnln(errPost)
 	}
+	if factory.WebUIConfig.Configuration.SendPebbleNotifications {
+		err := sendPebbleNotification("aetherproject.org/webconsole/networkslice/create")
+		if err != nil {
+			logger.ConfigLog.Warnf("sending Pebble notification failed: %s. continuing silently.", err.Error())
```
gab-arrobo, line 218:
```suggestion
			logger.ConfigLog.Warnf("sending Pebble notification failed: %s. continuing silently", err.Error())
```

proto/server/configEvtHandler.go
```diff
@@ -210,6 +212,34 @@ func handleNetworkSlicePost(configMsg *configmodels.ConfigMessage, subsUpdateCha
...
+	}
+	filter := bson.M{"slice-name": configMsg.SliceName}
+	err := dbadapter.CommonDBClient.RestfulAPIDeleteOne(sliceDataColl, filter)
+	if err != nil {
+		logger.DbLog.Warnln(err)
+	}
+	if factory.WebUIConfig.Configuration.SendPebbleNotifications {
+		err := sendPebbleNotification("aetherproject.org/webconsole/networkslice/delete")
+		if err != nil {
+			logger.ConfigLog.Warnf("sending Pebble notification failed: %s. continuing silently.", err.Error())
```
gab-arrobo, line 240:
```suggestion
			logger.ConfigLog.Warnf("sending Pebble notification failed: %s. continuing silently", err.Error())
```

proto/server/configEvtHandler.go
gab-arrobo: One more thing, in the previous code the writing event to channel happened only once. That is, delete device group (if needed), then delete slice (if needed) and then write event to channel. With the current changes, if there is a delete device group and delete slice, we would be writing  the event to channel twice. Is this the intention with these changes?

Gmerold: I don't think we're gonna write to the channel twice, because I removed the code which was originally responsible for doing that (lines 151-155). Now the handler functions are exclusively responsible for writing to the channel.

As for the order, it's a good question. I was thinking about the same. I decided to do it like that to be consistent with the rest of the handler functions. It's probably not critical, but I think it would make more sense to write to the channel after the operation is done; for all the handlers. How bout we take this to a separate PR?

proto/server/configEvtHandler.go
thakurajayL: Thanks

proto/server/configEvtHandler.go
gab-arrobo: > I don't think we're gonna write to the channel twice, because I removed the code which was originally responsible for doing that (lines 151-155). Now the handler functions are exclusively responsible for writing to the channel.

Yes, you removed the original write to channel code and put it inside each function (delete device group and delete slice). My point is that if delete device group is true and delete slice is true, you would be writing to channel twice. However, this was not the behavior before. That is, before if delete device group was true and delete slice was true, writing to channel happened only once. Do you agree?

gab-arrobo: > Please correct if I'm missing something, but I don't think we can be deleting a device group and a network slice at the same time. Especially after I added additional conditions to the `ifs`. If I'm not mistaken this would be 2 separate events.

I do not know whether it is doable to have both `delete device-group` and `delete slice` as a single operation. My comment was primary comparing of what the code is currently doing and the proposed changes in your PR.

@patriciareinoso, just in case, do you any comment/feedback about the changes proposed in this PR (especially, about whether a `delete group-device` and `delete slice` could be called in the same operation)? Thanks!

patriciareinoso: Following the execution from `webconsole/configapi/api_default.go`, my understanding is that Delete network slice and Delete device group operations are always handled as separate events. We will never have a `delete slice` and `delete device-group` at the same time in the  `configChannel` message. 
This being said, there is no change in behavior, we used to send 1 `Update5GSubscriberMsg` messages per event and we still do so after modifications.
I agree that using `if` instead of `else if` in lines 130 and 135 can be misleading.

gab-arrobo: @patriciareinoso, thanks for your feedback. The code before, first executed delete operation (device group or slice) and then writes to channel. With this PR, it happens the opposite, first write to channel and then execute delete operation. Any thoughts on this? Does this change have any side-effects? or does it not make any difference?

patriciareinoso: The order does not matter. The processing of the message in the channel is completely independent from the DB operations. I would keep the same order in the delete and post operations to make it consistent. 
I also think that extracting `handleNetworkSliceDelete` and `handleDeviceGroupDelete` methods is a good idea, it's more readable.

patriciareinoso (approved):

gab-arrobo (approved):
+1

PR #263, opened 2024-11-20
fix: swagger for user account creation

patriciareinoso:
The correct endpoint for user creation is `/config/v1/account/` and not `/config/v1/account/{username}`
This PR fixes the endpoint so that it has the correct URL in the swagger page.

## Before

image

## After 

image


gab-arrobo (approved):
+1

PR #278, opened 2025-01-08
feat: implement PUT for inventory UPF

patriciareinoso:
# Description

This PR implements the edition of UPFs

- Create an index over hostname for UPFs to guarantee unique hostnames in DB
- Modify POST operation to remove the target from path: now hostname and port are specified in the body of the request.
- We cannot create 2 UPFs with the same hostname.
- Add PUT operation:
   - If the UPF hostname does not exist it is created.
   - If the UPF hostname exists, it is modified.
- Add validation over PORT

Given that we can create a NS containing a UPF before or after the creation of a UPF object, the PUT and the POST UPF operations are propagated to the NS containing it.

Edition of UPF is propagated to Network slices containing a UPF with the given name using a PATCH operation to modify only the UPF field.

We use transactions to ensure that both DB updates are successfully executed or reverted.

The POST and PUT handlers do not use the config channel -> reduce complexity.


configapi/api_inventory.go
```diff
@@ -200,21 +198,145 @@ func GetUpfs(c *gin.Context) {
...
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to PUT UPF"})
+		return
+	}
+	c.JSON(http.StatusOK, gin.H{})
+}
+
+func postUpfOperation(upf configmodels.Upf, sc mongo.SessionContext) error {
+	filter := bson.M{"hostname": upf.Hostname}
+	upfDataBson := configmodels.ToBsonM(upf)
+	return dbadapter.CommonDBClient.RestfulAPIPostMany(configmodels.UpfDataColl, filter, []interface{}{upfDataBson})
```
patriciareinoso, line 294: need to modify the mongo API

configapi/api_inventory.go
```diff
@@ -291,3 +377,23 @@ func handleDeleteUpf(c *gin.Context) error {
 	logger.ConfigLog.Infof("successfully added UPF [%v] with delete_op to config channel", upfHostname)
 	return nil
 }
+
+func updateUpfInNetworkSlices(hostname string, patchJSON []byte, context context.Context) error {
```
patriciareinoso, line 381: I think that i can simplify this using the right filter

configapi/api_inventory.go
```diff
@@ -224,21 +220,145 @@ func GetUpfs(c *gin.Context) {
...
+}
+
+func putUpfOperation(upf configmodels.Upf, sc mongo.SessionContext) error {
+	filter := bson.M{"hostname": upf.Hostname}
+	upfDataBson := configmodels.ToBsonM(upf)
+	_, err := dbadapter.CommonDBClient.RestfulAPIPutOneWithContext(sc, configmodels.UpfDataColl, filter, upfDataBson)
+	return err
+}
+
+func handleUpfTransaction(ctx context.Context, upf configmodels.Upf, operation func(configmodels.Upf, mongo.SessionContext) error) error {
```
patriciareinoso, line 326: i can merge this with handleDeleteUpfTransaction

configapi/api_inventory.go
```diff
@@ -224,21 +220,109 @@ func GetUpfs(c *gin.Context) {
...
-// @Router       /config/v1/inventory/upf/{upf-hostname}  [post]
+// @Failure      500  {object}  nil  "Error creating UPF"
+// @Router       /config/v1/inventory/upf/  [post]
 func PostUpf(c *gin.Context) {
-	setInventoryCorsHeader(c)
-	if err := handlePostUpf(c); err == nil {
-		c.JSON(http.StatusOK, gin.H{})
-	} else {
-		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+	logger.WebUILog.Infoln("received a POST UPF request")
```
patriciareinoso, line 232: re add header

configapi/api_inventory.go
```diff
@@ -224,21 +219,113 @@ func GetUpfs(c *gin.Context) {
...
+	return []byte(fmt.Sprintf(`[
+		{
+			"op": "replace",
+			"path": "/site-info/upf",
+			"value": {
+				"upf-name": "%s",
+				"upf-port": "%s"
+			}
+		}
+	]`, upf.Hostname, upf.Port))
```
gatici, line 328: Why don't we prefer json encoding library to serialize the payload instead of manual string formattin using fmt.sprintf ?

configapi/api_inventory.go
```diff
@@ -377,7 +403,53 @@ func updateGnbInNetworkSlices(gnbName string, context context.Context) error {
 	return nil
 }
 
-func updateUpfInNetworkSlices(hostname string, patchJSON []byte, context context.Context) error {
+func postUpfOperation(sc mongo.SessionContext, upf configmodels.Upf) error {
+	filter := bson.M{"hostname": upf.Hostname}
+	upfDataBson := configmodels.ToBsonM(upf)
```
gatici, line 408: ToBsonM will return nil if there is any error occurs. We need to do a  nil check here before passing the nil data to post command.

configapi/api_inventory.go
```diff
@@ -377,7 +403,53 @@ func updateGnbInNetworkSlices(gnbName string, context context.Context) error {
...
+	return dbadapter.CommonDBClient.RestfulAPIPostManyWithContext(sc, configmodels.UpfDataColl, filter, []interface{}{upfDataBson})
+}
+
+func putUpfOperation(sc mongo.SessionContext, upf configmodels.Upf) error {
+	filter := bson.M{"hostname": upf.Hostname}
+	upfDataBson := configmodels.ToBsonM(upf)
```
gatici, line 414: We need to do a nil check for upfDataBson as if there is any error in ToBsonM, it will return nil.

gatici:
Looks good. A minor comments are added.

configapi/api_inventory.go
patriciareinoso: like this ?

gatici (approved):

ghislainbourgeois (approved):

ghislainbourgeois (approved):

PR #283, opened 2025-01-14
feat: propagate DG changes to all IMSIs

### LINKED ISSUE #284: Edit MBR upstream and downstream in Device Group does not propagate the values to subscribers in DB
## Description

the values of Maximum Bitrate (upstream and downstream)  of the Device Group are used in the `SubscribedUeAmbr` and `SessionAmbr `fields of the subscriber during creation. These fields are not modified in the subscriber when we edit them in a Device Group.

## How to reproduce

1. Create a Device Group
2. Create a subscriber associated to the DG.
3. Edit the Maximum Bitrate - Downstream and Maximum Bitrate - Upstream in the DG.
4. Check the content of the Subscriber 

```
 "AccessAndMobilitySubscriptionData": {
    "gpsis": [
      "msisdn-0900000000"
    ],
    "subscribedUeAmbr": {
      "uplink": "20 Mbps",
      "downlink": "200 Mbps"
    },
    "nssai": {
      "defaultSingleNssais": [
        {
          "sst": 1,
          "sd": "102030"
        }
      ],
      "singleNssais": [
        {
          "sst": 1,
          "sd": "102030"
        }
      ]
    }
  },
 "SessionManagementSubscriptionData": [
    {
      "singleNssai": {
        "sst": 1,
        "sd": "102030"
      },
      "dnnConfigurations": {
        "internet": {
          "pduSessionTypes": {
            "defaultSessionType": "IPV4",
            "allowedSessionTypes": [
              "IPV4"
            ]
          },
          "sscModes": {
            "defaultSscMode": "SSC_MODE_1",
            "allowedSscModes": [
              "SSC_MODE_2",
              "SSC_MODE_3"
            ]
          },
          "5gQosProfile": {
            "5qi": 9,
            "arp": {
              "priorityLevel": 8,
              "preemptCap": "",
              "preemptVuln": ""
            },
            "priorityLevel": 8
          },
          "sessionAmbr": {
            "uplink": "20 Mbps",
            "downlink": "200 Mbps"
          }
        }
      }
    }
  ],
```

dariofaccin:
This PR aims to fix #284 , the issue for which Access Mobility Data, Session Management Data, SMF Selection Data and Policy Data for subscribers are not updated when the Device Group is updated.

Before this PR, only subscribers that are added in the same API call of the Device Group update take the new information. This PR extends the refresh to all the subscribers currently belonging to the DG.

Current behaviour
A device group is created and its representation (only relevant part) in database is the following:
```
{
    [...]
    "imsis" : [ 
        "208930100007487"
    ],
    "ip-domain-expanded" : {
        "ue-dnn-qos" : {
            "dnn-mbr-uplink" : 20000000.0,
            "dnn-mbr-downlink" : 200000000.0,
            "bitrate-unit" : "bps",
            [...]
        },
        [...]
    }
}
```
The representation (only relevant part) of the subscriber associated to the device group is the following:
```
{
    [...]
    "subscribedUeAmbr" : {
        "uplink" : "20 Mbps",
        "downlink" : "200 Mbps"
    }
}
```
After updating the device group with new MBR data, the representation is the following:
```
{
    [...]
    "imsis" : [ 
        "208930100007487"
    ],
    "ip-domain-expanded" : {
        "ue-dnn-qos" : {
            "dnn-mbr-uplink" : 20000000.0,
            "dnn-mbr-downlink" : 300000000.0,
            "bitrate-unit" : "bps",
            [...]
        },
        [...]
    }
}
```
But subscriber data in all the collections is not updated:
```
{
    [...]
    "subscribedUeAmbr" : {
        "uplink" : "20 Mbps",
        "downlink" : "200 Mbps"
    }
}
```
The example is taken from the `amData` collection, but it applies also to `smData` and the others.


patriciareinoso:
I created an issue to better understand the problem and the purpose of the PR. Please add the reference to it in the PR description.  Would it be possible to capture a before and after of the DB content ? like this it would be clearer which fields we are talking about.

patriciareinoso:
What do you mean by "only subscribers that are added in the same API call of the Device Group update take the new information" ?  if the update api call does not include the list of subscribers it means that we are overwriting that list

patriciareinoso:
> What do you mean by "only subscribers that are added in the same API call of the Device Group update take the new information" ? if the update api call does not include the list of subscribers it means that we are overwriting that list

ok, looking at the code I think I understood. It seems like `getAddedImsisList` can be removed. But I am worried about the case in which you add a DG and add a reference to a Subscriber in the `imsis` list but you haven't created the actual Subscriber yet.

patriciareinoso (changes requested):
Oh I thought I did not have permission to request changes. Please see the comments I added

proto/server/configEvtHandler.go
```diff
@@ -616,14 +616,16 @@ func Config5GUpdateHandle(confChan chan *Update5GSubscriberMsg) {
...
-				for _, imsi := range aimsis {
-					dnn := confData.Msg.DevGroup.IpDomainExpanded.Dnn
-					updateAmPolicyData(imsi)
-					updateSmPolicyData(snssai, dnn, imsi)
-					updateAmProvisionedData(snssai, confData.Msg.DevGroup.IpDomainExpanded.UeDnnQos, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, imsi)
-					updateSmProvisionedData(snssai, confData.Msg.DevGroup.IpDomainExpanded.UeDnnQos, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, dnn, imsi)
-					updateSmfSelectionProviosionedData(snssai, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, dnn, imsi)
+				for _, imsi := range confData.Msg.DevGroup.Imsis {
+					/* update only if the imsi is provisioned */
+					if imsiData[imsi] != nil {
```
patriciareinoso, line 621: if `imsiData` is kept in memory. What happens if we add Subscriber A, B , C. and they exist in the `DevGroup.Imsis`. In the mean time the pod dies and then we update the MBR ?

patriciareinoso (changes requested):

proto/server/configEvtHandler.go
dariofaccin: This issue looks to me independent from this PR, as [the same logic is already in place](https://github.com/omec-project/webconsole/blob/3b49737b5d7d232e39b69a1f06f99a18987c84ef/proto/server/configEvtHandler.go#L360). I can try to fix it in this same PR or open a new one, I don't have a strong opinion.

patriciareinoso (changes requested):
if I try to build I get the error:
```
proto/server/clientEvtHandler.go:978:21: undefined: imsiData
```

proto/server/configEvtHandler.go
```diff
@@ -616,14 +641,16 @@ func Config5GUpdateHandle(confChan chan *Update5GSubscriberMsg) {
...
 
-				aimsis := getAddedImsisList(confData.Msg.DevGroup, confData.PrevDevGroup)
-				for _, imsi := range aimsis {
-					dnn := confData.Msg.DevGroup.IpDomainExpanded.Dnn
-					updateAmPolicyData(imsi)
-					updateSmPolicyData(snssai, dnn, imsi)
-					updateAmProvisionedData(snssai, confData.Msg.DevGroup.IpDomainExpanded.UeDnnQos, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, imsi)
-					updateSmProvisionedData(snssai, confData.Msg.DevGroup.IpDomainExpanded.UeDnnQos, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, dnn, imsi)
-					updateSmfSelectionProviosionedData(snssai, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, dnn, imsi)
+				for _, imsi := range confData.Msg.DevGroup.Imsis {
```
patriciareinoso, line 644: This method is shared between the DELETE and the POST operation. In the delete operation confData.Msg.DevGroup is nil, so the program crashes if we try to delete a DG.

patriciareinoso (changes requested):

patriciareinoso (changes requested):
These modifications change the initial behavior of the following workflow in the NMS:

1. Create a NS
2. DG is automatically created
3. Create a Subscriber

# Before
`subscriptionData.provisionedData.amData`, `subscriptionData.provisionedData.smData`, `subscriptionData.provisionedData.smfSelectionSubscriptionData` are filled

## Now
`subscriptionData.provisionedData.amData`, `subscriptionData.provisionedData.smData`, `subscriptionData.provisionedData.smfSelectionSubscriptionData` are not filled

proto/server/configEvtHandler.go
```diff
@@ -607,6 +631,7 @@ func Config5GUpdateHandle(confChan chan *Update5GSubscriberMsg) {
 			/* is this devicegroup part of any existing slice */
 			slice := isDeviceGroupExistInSlice(confData)
 			if slice != nil {
+				provisionedSubscribers := getProvisionedSubscribers()
```
patriciareinoso, line 634: i would make this in line 646: after the if and before the loop. In case of delete operation we would unnecessarily retrieving the list of subscribers

proto/server/configEvtHandler.go
```diff
@@ -402,6 +396,36 @@ func getDeletedImsisList(group, prevGroup *configmodels.DeviceGroups) (dimsis []
...
+		ueId, ok := rawProvisionedSubscriber["ueId"].(string)
+		if !ok {
+			logger.DbLog.Warnf("cannot retrieve ueId for subscriber: %v", rawProvisionedSubscriber)
+		}
+		provisionedSubscribers = append(provisionedSubscribers, ueId)
+	}
+	return provisionedSubscribers
+}
+
+func getSubscriberAuthDataByUeId(ueId string) *models.AuthenticationSubscription {
```
patriciareinoso, line 415: i would move this function to clientEvtHandler.go since it is only used there

proto/server/configEvtHandler.go
```diff
@@ -402,6 +396,36 @@ func getDeletedImsisList(group, prevGroup *configmodels.DeviceGroups) (dimsis []
 	return
 }
 
+func getProvisionedSubscribers() []string {
+	rawProvisionedSubscribers, errGetMany := dbadapter.AuthDBClient.RestfulAPIGetMany(authSubsDataColl, nil)
+	if errGetMany != nil {
+		logger.DbLog.Warnln(errGetMany)
```
patriciareinoso, line 402: we can inmmediately return here

proto/server/configEvtHandler.go
```diff
@@ -402,6 +396,36 @@ func getDeletedImsisList(group, prevGroup *configmodels.DeviceGroups) (dimsis []
...
+		provisionedSubscribers = append(provisionedSubscribers, ueId)
+	}
+	return provisionedSubscribers
+}
+
+func getSubscriberAuthDataByUeId(ueId string) *models.AuthenticationSubscription {
+	filter := bson.M{"ueId": ueId}
+	subscriberAuthDataInterface, errGetOne := dbadapter.AuthDBClient.RestfulAPIGetOne(authSubsDataColl, filter)
+	if errGetOne != nil {
+		logger.DbLog.Warnln(errGetOne)
```
patriciareinoso, line 419: we can immediately return here

proto/server/configEvtHandler.go
```diff
@@ -355,9 +348,10 @@ func getAddedImsisList(group, prevGroup *configmodels.DeviceGroups) (aimsis []st
 	if group == nil {
 		return
 	}
+	provisionedSubscribers := getProvisionedSubscribers()
```
patriciareinoso, line 351: i am concerned that here you are using the information in `authSubsDataColl` but that table is only filled up in `factory.WebUIConfig.Configuration.Mode5G`  Do you know what happens in all the function in `clientEvtHandler` that use `getAddedImsisList`? are all of them used when `factory.WebUIConfig.Configuration.Mode5G` ?

patriciareinoso (changes requested):

proto/server/configEvtHandler.go
dariofaccin: In `clientEvtHandler` we are not istantiating the DBClient, therefore moving this function there will increase the imports and in general the complexity. In the client handler we are also already importing `getAddedImsisList()` func from `configEvtHandler`

proto/server/configEvtHandler.go
dariofaccin: In `clientEvtHandler` when `5GMode` is disabled there are other functions (in this case, `postConfigHss`) that take care of the subscribers and these are using another functions (in this case, `addedImsis()` and `deletedImsis()`).

proto/server/configEvtHandler.go
gatici: Agree, if we continue `rawProvisionedSubscribers` might be uninitialized and return an `nil pointer dereference` in line 405.

proto/server/configEvtHandler.go
```diff
@@ -402,6 +396,36 @@ func getDeletedImsisList(group, prevGroup *configmodels.DeviceGroups) (dimsis []
...
+	rawProvisionedSubscribers, errGetMany := dbadapter.AuthDBClient.RestfulAPIGetMany(authSubsDataColl, nil)
+	if errGetMany != nil {
+		logger.DbLog.Warnln(errGetMany)
+	}
+	var provisionedSubscribers []string
+	for _, rawProvisionedSubscriber := range rawProvisionedSubscribers {
+		ueId, ok := rawProvisionedSubscriber["ueId"].(string)
+		if !ok {
+			logger.DbLog.Warnf("cannot retrieve ueId for subscriber: %v", rawProvisionedSubscriber)
+		}
```
gatici, line 409:
```suggestion
		if !ok {
			logger.DbLog.Warnf("cannot retrieve ueId for subscriber: %v", rawProvisionedSubscriber)
			continue
		}
```
If we can not get a proper ueId, we can skip it.

proto/server/configEvtHandler.go
gatici: We can return an empty slice.

proto/server/configEvtHandler.go
```diff
@@ -402,6 +396,36 @@ func getDeletedImsisList(group, prevGroup *configmodels.DeviceGroups) (dimsis []
...
+	filter := bson.M{"ueId": ueId}
+	subscriberAuthDataInterface, errGetOne := dbadapter.AuthDBClient.RestfulAPIGetOne(authSubsDataColl, filter)
+	if errGetOne != nil {
+		logger.DbLog.Warnln(errGetOne)
+	}
+	var subscriberAuthData models.AuthenticationSubscription
+	err := json.Unmarshal(configmodels.MapToByte(subscriberAuthDataInterface), &subscriberAuthData)
+	if err != nil {
+		logger.DbLog.Errorf("could not unmarshall subscriber %v", subscriberAuthDataInterface)
+	}
```
gatici, line 425:
```suggestion
	if err != nil {
		logger.DbLog.Errorf("could not unmarshall subscriber %v", subscriberAuthDataInterface)
		return nil
	}
```
We can return here.

proto/server/configEvtHandler.go
```diff
@@ -616,14 +641,19 @@ func Config5GUpdateHandle(confChan chan *Update5GSubscriberMsg) {
 					Sst: int32(sVal),
```
gatici, line 641: We shoud return after line 637. If strconv.ParseUint fails, we can not calculate snssai properly.

proto/server/configEvtHandler.go
```diff
@@ -616,14 +641,19 @@ func Config5GUpdateHandle(confChan chan *Update5GSubscriberMsg) {
...
-				aimsis := getAddedImsisList(confData.Msg.DevGroup, confData.PrevDevGroup)
-				for _, imsi := range aimsis {
-					dnn := confData.Msg.DevGroup.IpDomainExpanded.Dnn
-					updateAmPolicyData(imsi)
-					updateSmPolicyData(snssai, dnn, imsi)
-					updateAmProvisionedData(snssai, confData.Msg.DevGroup.IpDomainExpanded.UeDnnQos, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, imsi)
-					updateSmProvisionedData(snssai, confData.Msg.DevGroup.IpDomainExpanded.UeDnnQos, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, dnn, imsi)
-					updateSmfSelectionProviosionedData(snssai, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, dnn, imsi)
+				/* skip delete case */
+				if confData.Msg.DevGroup != nil {
```
gatici, line 645: We can move the first if clause outside of this block to increase readability.

  if confData.Msg.DevGroup == nil {
      logger.DbLog.Warn("DevGroup is nil in confData.Msg")
      return
  }

gatici (changes requested):

proto/server/configEvtHandler.go
dariofaccin: I don't see the point of adding `return nil` here: in line 427 we are returning the pointer (`return &subscriberAuthData`) which will already be `nil` if the unmarshalling failed.

proto/server/configEvtHandler.go
dariofaccin: If `DevGroup` is nil we want to remove the subscribers data from the collections, but not update them. I don't fully understand the comment however.

proto/server/configEvtHandler.go
gatici: In the project L978 this function is called like below:
```
				authSubsData := getSubscriberAuthDataByUeId("imsi-" + imsi)
				if authSubsData == nil {
					client.clientLog.Infoln("SIM card details not found for IMSI ", imsi)
					continue
				}
```

They do a nil check to skip or continue to execution.
If `json.Unmarshal` fails, &subscriberAuthData pointer  will keep the location of an empty AuthenticationSubscription  struct looking like:
{Opc:{...} PermanentKey:{...} SequenceNumber: ""}
And it will not be `nil`. Execution continue and In L983 `config.Opc = authSubsData.Opc.OpcValue`  will cause a panic as `OpcValue` is not reachable.
Hence, instead of returning a pointer of empty struct, we can return nil in L428.

proto/server/configEvtHandler.go
gatici: Let's talk about it. I will ping you.

gatici: Please disregard my comment.

gatici (approved):

patriciareinoso (approved):

gab-arrobo:
@dariofaccin, I guess you have tested these changes, correct? I am asking this because I was running a test and I do not see these changes being used, or am I missing something?

dariofaccin:
@gab-arrobo, these changes are used to update the subscriber data of a device group whenever the device group is updated. To see the changes the flow is the following:
1. create a device group (and network slice if needed)
2. create a subscriber associated to such device group
3. edit the device group
4. changes are reflected in the subscriber data (collections and data returned by `GET /subscribers/{imsi}`)

Let me know, thanks

proto/server/clientEvtHandler.go
```diff
@@ -975,7 +975,7 @@ func postConfigHss(client *clientNF, lastDevGroup *configmodels.DeviceGroups, la
 				}
 				config.StartImsi = uint64(num)
 				config.EndImsi = uint64(num)
-				authSubsData := imsiData[imsi]
+				authSubsData := getSubscriberAuthDataByUeId("imsi-" + imsi)
```
patriciareinoso, line 978: this piece of code is executed by the 4G configuration , in which case the `authSubsDataColl` used in `getSubscriberAuthDataByUeId` will not be filled.

And it's not 100% clear to me if `getAddedImsisList` if only used by 5G code since the code is shared between both configurations

patriciareinoso (changes requested):

thakurajayL:
Is this ready for review? I see its in draft state.

proto/server/configEvtHandler.go
```diff
@@ -535,20 +535,26 @@ func Config5GUpdateHandle(confChan chan *Update5GSubscriberMsg) {
...
```
patriciareinoso, line 546: we can use the type of operation in here to differentiate the POST from DELETE. It would be more readable than having to understand in which case the confData.Msg.DevGroup != nil and in which it is not.

patriciareinoso (commented):
LGTM, just a small comment for readability

proto/server/configEvtHandler.go
```diff
@@ -535,20 +535,26 @@ func Config5GUpdateHandle(confChan chan *Update5GSubscriberMsg) {
...
+						/* update only if the imsi is provisioned */
+						if subscriberAuthData.SubscriberAuthenticationDataGet("imsi-"+imsi) != nil {
+							dnn := confData.Msg.DevGroup.IpDomainExpanded.Dnn
+							updateAmPolicyData(imsi)
+							updateSmPolicyData(snssai, dnn, imsi)
+							updateAmProvisionedData(snssai, confData.Msg.DevGroup.IpDomainExpanded.UeDnnQos, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, imsi)
+							updateSmProvisionedData(snssai, confData.Msg.DevGroup.IpDomainExpanded.UeDnnQos, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, dnn, imsi)
+							updateSmfSelectionProviosionedData(snssai, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, dnn, imsi)
+						}
+					}
```
ghislainbourgeois, line 557: The previous code, if I understand it correctly, was meant to provision the new IMSIs after a device group was updated (potentially just by adding new IMSIs). The new code seems to only run the update for existing IMSIs. I think this is incorrect.

What I think needs to happen here is that:

1. IMSIs that are in the new device group but were not there before are added
2. IMSIs that were already present are updated

I think removing the `if` like so would accomplish this in the simplest way. WDYT?

```suggestion
				/* skip delete case */
				if confData.Msg.MsgMethod != configmodels.Delete_op {
					for _, imsi := range confData.Msg.DevGroup.Imsis {
						dnn := confData.Msg.DevGroup.IpDomainExpanded.Dnn
						updateAmPolicyData(imsi)
						updateSmPolicyData(snssai, dnn, imsi)
						updateAmProvisionedData(snssai, confData.Msg.DevGroup.IpDomainExpanded.UeDnnQos, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, imsi)
						updateSmProvisionedData(snssai, confData.Msg.DevGroup.IpDomainExpanded.UeDnnQos, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, dnn, imsi)
						updateSmfSelectionProviosionedData(snssai, slice.SiteInfo.Plmn.Mcc, slice.SiteInfo.Plmn.Mnc, dnn, imsi)
					}
```

ghislainbourgeois (changes requested):
I have a concern for what happens when a subscriber is added in a device group POST but is not already in `subscriberAuthData`. I want to make sure this change will not break the other users of this API.

ghislainbourgeois (approved):

PR #299, opened 2025-02-04
feat: propagate Subscriber deletion to Device Groups

dariofaccin:
# Issue
When we DELETE a subscriber, information is removed from `subscriptionData.provisionedData.amData` and `subscriptionData.authenticationData.authenticationSubscription`. However, if a device group contains a reference to this subscriber in `imsis` list, the reference is not removed.

# Description
This PR propagates the DELETE subscriber operation to all the device groups that contains the deletedsubscriber. This way, the removal of the subscriber is done in a single operation, no need to manually update the device groups to remove the dangling reference anymore.
This PR improves the response when the list of subscribers is empty: now it returns an empty list (`[]`) instead of `null`.
This PR also improves the response in case of subscriber deletion failure: now it returns 500 Internal Server Error.

# How the code works
### Webconsole behavior on Subscriber POST before this PR
When a subscriber is created (POST), it creates entries in:
```
subscriptionData.authenticationData.authenticationSubscription
subscriptionData.provisionedData.amData
```
### Webconsole behavior on Subscriber DELETE before this PR
When a subscriber is deleted (DELETE) , it removes entries for each subscriber in:
```
subscriptionData.authenticationData.authenticationSubscription
subscriptionData.provisionedData.amData
```
# This PR
* retrieves all the device groups where the subscriber to be deleted is present
* creates a new (in memory) configuration for all the device groups on previous point, where the subscriber is removed from `imsis`
* sends a message in the config channel for the subscriber removal
* sends a message in the config channel for the device groups update

# How to reproduce
1. Create a subscriber, a device group containing such subscriber and a network slice containing such device group
2. Check the DB entries for that device group in `webconsoleData.snapshots.devGroupData` (`imsis` list)
3. Delete the subscriber

### What you should see
1. The subscriber reference is removed from the device group(s) `imsis` list
2. The subscriber information is removed from `subscriptionData.authenticationData.authenticationSubscription`
3. The subscriber information with the PLMN belonging to the modified device groups is removed from:
`subscriptionData.provisionedData.amData`
`subscriptionData.provisionedData.smData`
`subscriptionData.provisionedData.smfSelectionSubscriptionData`
`policyData.ues.amData`
`policyData.ues.smData`


configapi/api_sub_config_test.go
```diff
@@ -0,0 +1,281 @@
...
+					t.Errorf("expected MsgMethod %+v, but got %+v", tc.expectedMessage.MsgMethod, msg.MsgMethod)
+				}
+				if tc.expectedMessage.AuthSubData != nil {
+					if msg.AuthSubData == nil {
+						t.Errorf("expected AuthSubData %+v, but got nil", tc.expectedMessage.AuthSubData)
+					}
+					if tc.expectedMessage.Imsi != msg.Imsi {
+						t.Errorf("expected IMSI %+v, but got %+v", tc.expectedMessage.Imsi, msg.Imsi)
+					}
+				}
```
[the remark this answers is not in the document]
patriciareinoso: you can also add a check to verify that the `AuthSubData` in the received message is nil. Or maybe use `reflect.DeepEqual`  to check the whole message in one step instead of checking each field

patriciareinoso:
Please add to the description the fact that you are changing the behavior of the getter to return [ ] instead of null

configapi/api_sub_config_test.go
```diff
@@ -0,0 +1,354 @@
...
+	testCases := []struct {
+		name            string
+		route           string
+		inputData       string
+		expectedMessage configmodels.ConfigMessage
+	}{
+		{
+			name:      "Create a new subscriber success",
+			route:     "/api/subscriber/imsi-208930100007487",
+			inputData: `{"plmnId":"12345", "opc":"8e27b6af0e692e750f32667a3b14605d","key":"8baf473f2f8fd09487cccbd7097c6862", "sequenceNumber":"16f3b3f70fc2"}`,
```
patriciareinoso, line 151:
```suggestion
			inputData: `{"plmnID":"12345", "opc":"8e27b6af0e692e750f32667a3b14605d","key":"8baf473f2f8fd09487cccbd7097c6862", "sequenceNumber":"16f3b3f70fc2"}`,
```

configapi/api_sub_config_test.go
```diff
@@ -0,0 +1,354 @@
...
+				if msg.MsgMethod != tc.expectedMessage.MsgMethod {
+					t.Errorf("expected MsgMethod %+v, but got %+v", tc.expectedMessage.MsgMethod, msg.MsgMethod)
+				}
+				if tc.expectedMessage.AuthSubData != nil {
+					if msg.AuthSubData == nil {
+						t.Errorf("expected AuthSubData %+v, but got nil", tc.expectedMessage.AuthSubData)
+					}
+					if tc.expectedMessage.Imsi != msg.Imsi {
+						t.Errorf("expected IMSI %+v, but got %+v", tc.expectedMessage.Imsi, msg.Imsi)
+					}
```
patriciareinoso, line 300: this part of code is never executed because `tc.expectedMessage.AuthSubData != nil` condition is never true.  

it's not recommended to have complicated logic in the checks. 
We need to test that in the first tc only one message is sent, and in the second tc we need to check that 2 messages are sent, if the logic becomes complicated it's better to separate en 2 tests.
Complicated logic in tests is error prone, and hard to read

patriciareinoso (changes requested):

configapi/api_sub_config_test.go
```diff
@@ -0,0 +1,354 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2024 Canonical Ltd.
```
patriciareinoso, line 2:
```suggestion
// Copyright 2025 Canonical Ltd.
```

configapi/api_sub_config.go
```diff
@@ -494,6 +501,39 @@ func DeleteSubscriberByID(c *gin.Context) {
 	logger.WebUILog.Infoln("Delete Subscriber Data complete")
 }
 
+func updateSubscriberInDeviceGroups(imsi string) {
+	filterByImsi := bson.M{
+		"imsis": imsi,
+	}
+	rawDeviceGroups, err := dbadapter.CommonDBClient.RestfulAPIGetMany(devGroupDataColl, filterByImsi)
```
gatici, line 508: In RestfulAPIGetMany `var resultArray []map[string]interface{}` initial value is nil. According to `cur, err := collection.Find(ctx, filter)`, if filter does not match it can stay as nil and it can return rawDeviceGroups  as nil  and err as nil. 
In line 509, as we only check the err is nil or not, execution continue. 
On line 513,  iterating over rawDeviceGroups which is nil may cause panic.

configapi/api_sub_config.go
```diff
@@ -494,6 +501,39 @@ func DeleteSubscriberByID(c *gin.Context) {
...
+	if err != nil {
+		logger.WebUILog.Errorf("failed to fetch device groups: %w", err)
+		return
+	}
+	for _, rawDeviceGroup := range rawDeviceGroups {
+		var deviceGroup configmodels.DeviceGroups
+		if err = json.Unmarshal(configmodels.MapToByte(rawDeviceGroup), &deviceGroup); err != nil {
+			logger.WebUILog.Errorf("error unmarshaling device group: %v", err)
```
gatici, line 516: we can improve logging here by improving device group name.

configapi/api_sub_config.go
```diff
@@ -485,6 +489,9 @@ func DeleteSubscriberByID(c *gin.Context) {
 
 	c.JSON(http.StatusNoContent, gin.H{})
 
+	imsi := strings.TrimPrefix(ueId, "imsi-")
+	updateSubscriberInDeviceGroups(imsi)
```
gatici, line 493: According to line 490, before doing any operation, `DeleteSubscriberByID` function send HTTP  204  no content response, then it calls updateSubscriberInDeviceGroups(imsi) method. I saw two problems here:
1. Order: response is sent without checking the success of device group update
2. Error propagation: updateSubscriberInDeviceGroups log the errors but does not propagate them to caller method. The caller always report a successful operation regardless from what happened in the updateSubscriberInDeviceGroups.

configapi/api_sub_config.go
```diff
@@ -494,6 +501,39 @@ func DeleteSubscriberByID(c *gin.Context) {
 	logger.WebUILog.Infoln("Delete Subscriber Data complete")
 }
 
+func updateSubscriberInDeviceGroups(imsi string) {
+	filterByImsi := bson.M{
+		"imsis": imsi,
```
gatici, line 506: I suggest to validate the existence of imsi here.

configapi/api_sub_config.go
dariofaccin: Initial value of `var resultArray []map[string]interface{}` is not `nil` but `[]`. Actually, there is a test (`TestSubscriberDeleteNoDeviceGroup`) where the IMSI to be deleted does not belong to any device group and it does not panic. From the debugger: `[]map[string]interface {}=[]`

configapi/api_sub_config.go
dariofaccin: At this point we are not sure that the `DeviceGroupName` field is available in the object we are trying to unmarshal: logging the raw object could simplify the debugging operations.

configapi/api_sub_config.go
gatici: You are right. Please discard this comment.

configapi/api_sub_config.go
gatici: OK

gab-arrobo:
@dariofaccin, you need to sing every commit you make (-s), otherwise, the DCO will fail

configapi/api_sub_config_test.go
```diff
@@ -0,0 +1,361 @@
...
+			if expectedMessage.MsgMethod != msg.MsgMethod {
+				t.Errorf("expected MsgMethod %+v, but got %+v", expectedMessage.MsgMethod, msg.MsgMethod)
+			}
+			if expectedMessage.Imsi != msg.Imsi {
+				t.Errorf("expected IMSI %+v, but got %+v", expectedMessage.Imsi, msg.Imsi)
+			}
+		default:
+			t.Error("expected message in configChannel, but none received")
+	}
+}
```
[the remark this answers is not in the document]
gatici: We need to try to read message again from the channel and msg should not exist.

configapi/api_sub_config_test.go
```diff
@@ -0,0 +1,415 @@
...
+	default:
+	}
+}
+
+func TestSubscriberDeleteSuccessWithDeviceGroup(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+	router := gin.Default()
+	AddApiService(router)
+	dbAdapter := &MockMongoClientDeviceGroupsWithSubscriber{}
+	dbadapter.CommonDBClient = dbAdapter
```
gatici, line 313: Generally speaking, we are accessing the `dbadapter.CommonDBClient` and `configChannel` global variables and changing their states in the tests. I saw some tests doing it but some are not. All the tests that accessing those global variables need to save their original states and after test, we need to restore them.
For example:

	// Save original state
	origDBClient := dbadapter.CommonDBClient
	origConfigChannel := configChannel

	// Restore them after the test finishes
	defer func() {
		dbadapter.CommonDBClient = origDBClient
		configChannel = origConfigChannel
	}()

	// Set up a mock database adapter
	dbAdapter := &MockMongoClientDeviceGroupsWithSubscriber{}
	dbadapter.CommonDBClient = dbAdapter
	
	// Create a new llocal config channel for the test
	configChannel = make(chan *configmodels.ConfigMessage, 2)
	
	...

gatici (commented):
Looks good.

patriciareinoso (approved):

configapi/api_sub_config.go
```diff
@@ -268,9 +269,12 @@ func GetSubscribers(c *gin.Context) {
 	logger.WebUILog.Infoln("Get All Subscribers List")
 
 	var subsList []configmodels.SubsListIE
+	subsList = make([]configmodels.SubsListIE, 0)
```
ghislainbourgeois, line 272:
```suggestion
	subsList := make([]configmodels.SubsListIE, 0)
```

configapi/api_sub_config.go
```diff
@@ -268,9 +269,12 @@ func GetSubscribers(c *gin.Context) {
...
 	amDataList, errGetMany := dbadapter.CommonDBClient.RestfulAPIGetMany(amDataColl, bson.M{})
 	if errGetMany != nil {
-		logger.DbLog.Warnln(errGetMany)
+		logger.DbLog.Errorw("failed to retrieve subscribers list", "error", errGetMany)
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to retrieve subscribers list"})
+		return
```
ghislainbourgeois, line 277: @gab-arrobo, this technically changes the API, but in a way that I think should not cause any issues. What do you think?

ghislainbourgeois (changes requested):
I have a small suggestion, and want to validate with @gab-arrobo that the small API change will not cause issues.

configapi/api_sub_config.go
gab-arrobo: @ghislainbourgeois, I have not had a chance to (re)create the APIs using code generator. So, I think we can merge this PR when your requested changes have been addressed.

thakurajayL (approved):

configapi/api_sub_config.go
dariofaccin: @gab-arrobo when you merge the PR and squash the commits, could please just keep "
feat: propagate Subscriber deletion to Device Groups" as a commit message? There is a lot of useless back and forth in the commit history. Thanks.

gab-arrobo (approved):
+1

configapi/api_sub_config.go
gab-arrobo: @ghislainbourgeois, I cannot merge the PR because you requested changes and GitHub requires for you approve the PR for it to be merged. So, after approving the PR and if you are planning to merge the PR, please take into account @dariofaccin's previous comment about the commit message

ghislainbourgeois (approved):

PR #310, opened 2025-02-20
feat: fully separate subscriber authentication handling between 4G and 5G

dariofaccin:
# Issue
When we create a subscriber, authentication data is handled differently whether `Mode5G` is enabled or not.
In 4G mode, authentication data is stored in a map in memory (`imsiData`) and its retrieval is done accessing such map.
In 5G mode, authentication data is stored both in memory (`imsiData`) and database (`subscriptionData.authenticationData.authenticationSubscription` collection). Its retrieval is however done accessing only the in-memory map. In this situation, a pod restart may cause the in-memory map to be reset while the database remains populated and therefore a state inconsistency.

# Description
This PR adds a new interface, named `SubscriberAuthenticationData`, which defines methods to retrieve, create/update and delete authentication data for a subscriber.
This PR implements it in two ways:
* `DatabaseSubscriberAuthenticationData`: this implementation handles the authentication data for 5G;
* `MemorySubscriberAuthenticationData`: this implementation handles the authentication data for 4G.

# How the code works (5G)
### Webconsole behavior on Subscriber POST before this PR (5G)
When a subscriber is created (POST), it creates entries in:
```
subscriptionData.authenticationData.authenticationSubscription [database collection]
imsiData [in-memory map]
```
### Webconsole behavior on Subscriber DELETE before this PR
When a subscriber is deleted (DELETE), it removes entries from:
```
subscriptionData.authenticationData.authenticationSubscription [database collection]
```
Authentication data for the subscriber is not removed from `imsiData`.

# How the code works (4G)
### Webconsole behavior on Subscriber POST before this PR (5G)
When a subscriber is created (POST), it creates entries in:
```
imsiData [in-memory map]
```
### Webconsole behavior on Subscriber DELETE before this PR
When a subscriber is deleted (DELETE), authentication data for the subscriber is not removed from `imsiData`.

# This PR
* instantiates the appropriate `SubscriberAuthenticationData` depending on `factory.WebUIConfig.Configuration.Mode5G`
* populates only the database (or only memory) on Subscriber POST
* fully removes the authentication data from the database (or memory) on Subscriber DELETE


proto/server/subscriber.go
```diff
@@ -0,0 +1,105 @@
...
+	SubscriberAuthenticationData
+}
+
+func (subscriberAuthData DatabaseSubscriberAuthenticationData) SubscriberAuthenticationDataGet(imsi string) (authSubData *models.AuthenticationSubscription) {
+	filter := bson.M{"ueId": imsi}
+	authSubDataInterface, errGetOne := dbadapter.CommonDBClient.RestfulAPIGetOne(authSubsDataColl, filter)
+	if errGetOne != nil {
+		logger.DbLog.Warnln(errGetOne)
+	}
+	err := json.Unmarshal(configmodels.MapToByte(authSubDataInterface), &authSubData)
```
ghislainbourgeois, line 32:
```suggestion
	authSubDataInterface, err := dbadapter.CommonDBClient.RestfulAPIGetOne(authSubsDataColl, filter)
	if err != nil {
		logger.DbLog.Warnln(err)
	}
	err = json.Unmarshal(configmodels.MapToByte(authSubDataInterface), &authSubData)
```

proto/server/subscriber.go
```diff
@@ -0,0 +1,105 @@
...
+	filter := bson.M{"ueId": imsi}
+	basicAmData := map[string]interface{}{
+		"ueId": imsi,
+	}
+	logger.WebUILog.Debugf("insert/update authentication subscription in amData collection: %v", imsi)
+	basicDataBson := configmodels.ToBsonM(basicAmData)
+	_, errPost := dbadapter.CommonDBClient.RestfulAPIPost(amDataColl, filter, basicDataBson)
+	if errPost != nil {
+		logger.DbLog.Warnln(errPost)
+	}
```
ghislainbourgeois, line 93:
```suggestion
	_, err := dbadapter.CommonDBClient.RestfulAPIPost(amDataColl, filter, basicDataBson)
	if err != nil {
		logger.DbLog.Warnln(err)
	}
```

ghislainbourgeois (changes requested):
I think we should take the opportunity to fix the different names for error variables.

I also think it might be better to just use the database. As far as I am aware, 4G is not really supported anymore, so I don't think it makes sense to make the code more complex for this. @gab-arrobo 
@thakurajayL, any thoughts on this?

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +440,189 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+	}
+
+	var AuthSubResult map[string]interface{} = postData[0]["data"].(map[string]interface{})
+	var AmDataResult map[string]interface{} = postData[0]["data"].(map[string]interface{})
+	if AuthSubResult["ueId"] != ueId {
+		t.Errorf("Expected ueId %v, got %v", ueId, AuthSubResult["ueId"])
+	}
+	if AmDataResult["ueId"] != ueId {
+		t.Errorf("Expected ueId %v, got %v", ueId, AmDataResult["ueId"])
+	}
```
patriciareinoso, line 500: can we check more than just the ueId here?  using reflect should be easy to check the whole struct

thakurajayL:
I will review this today

thakurajayL:
Generic comment - we need not improve 4G behavior. Let's it it as is. If you want to move around code and keep the 4G behavior as is then its acceptable.

proto/server/configEvtHandler.go
```diff
@@ -64,17 +65,13 @@ func configHandler(configMsgChan chan *configmodels.ConfigMessage, configReceive
...
-			rwLock.Lock()
-			imsiData[imsiVal] = configMsg.AuthSubData
-			rwLock.Unlock()
-			logger.ConfigLog.Infof("received Imsi [%v] configuration from config channel", configMsg.Imsi)
-			handleSubscriberPost(configMsg)
-			if factory.WebUIConfig.Configuration.Mode5G {
-				var configUMsg Update5GSubscriberMsg
-				configUMsg.Msg = configMsg
-				subsUpdateChan <- &configUMsg
+			logger.ConfigLog.Infoln("received imsi from config channel: ", imsiVal)
```
gab-arrobo, line 68:
```suggestion
			logger.ConfigLog.Infoln("received imsi from config channel:", imsiVal)
```

proto/server/subscriber.go
```diff
@@ -0,0 +1,105 @@
...
+
+type DatabaseSubscriberAuthenticationData struct {
+	SubscriberAuthenticationData
+}
+
+func (subscriberAuthData DatabaseSubscriberAuthenticationData) SubscriberAuthenticationDataGet(imsi string) (authSubData *models.AuthenticationSubscription) {
+	filter := bson.M{"ueId": imsi}
+	authSubDataInterface, err := dbadapter.CommonDBClient.RestfulAPIGetOne(authSubsDataColl, filter)
+	if err != nil {
+		logger.DbLog.Warnln(err)
```
patriciareinoso, line 30: we should return after the error

proto/server/subscriber.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	err = json.Unmarshal(configmodels.MapToByte(authSubDataInterface), &authSubData)
+	if err != nil {
+		logger.DbLog.Errorf("could not unmarshall subscriber %v", authSubDataInterface)
+	}
+	return authSubData
```
patriciareinoso, line 36: what is the value of `authSubData` if  error go over line 34?

proto/server/subscriber.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (subscriberAuthData DatabaseSubscriberAuthenticationData) SubscriberAuthenticationDataCreate(imsi string, authSubData *models.AuthenticationSubscription) {
```
patriciareinoso, line 39: i think that in this function we can use a mongoDB transaction to ensure that either both tables are updated or none

proto/server/subscriber.go
```diff
@@ -0,0 +1,105 @@
...
+		"ueId": imsi,
+	}
+	basicDataBson := configmodels.ToBsonM(basicAmData)
+	_, err = dbadapter.CommonDBClient.RestfulAPIPost(amDataColl, filter, basicDataBson)
+	if err != nil {
+		logger.DbLog.Warnln(err)
+	}
+}
+
+func (subscriberAuthData DatabaseSubscriberAuthenticationData) SubscriberAuthenticationDataDelete(imsi string) {
```
patriciareinoso, line 59: i think that in this function we can use a mongoDB transaction to ensure that either both tables are updated or none

proto/server/configEvtHandler.go
```diff
@@ -39,9 +39,10 @@ type Update5GSubscriberMsg struct {
 }
 
 var (
-	execCommand = exec.Command
-	imsiData    map[string]*models.AuthenticationSubscription
-	rwLock      sync.RWMutex
+	execCommand        = exec.Command
+	imsiData           map[string]*models.AuthenticationSubscription
```
patriciareinoso, line 43: do we need to keep the `imsiData` ? Could it be only associated to the `MemorySubscriberAuthenticationData`

proto/server/configEvtHandler.go
```diff
@@ -294,7 +289,7 @@ func getAddedImsisList(group, prevGroup *configmodels.DeviceGroups) (aimsis []st
 	}
 	for _, imsi := range group.Imsis {
 		if prevGroup == nil {
-			if imsiData[imsi] != nil {
```
patriciareinoso, line 297: how does this function work? because we are calling it using the imsi list in the device groups which do not have the `imsi-` prefix and in SubscriberAuthenticationDataGet we use the  `imsi-` prefix

patriciareinoso (changes requested):

proto/server/configEvtHandler.go
patriciareinoso: i was thinking about something like 
```
type MemorySubscriberAuthenticationData struct {
	SubscriberAuthenticationData
        imsiData
}
```
but i don't even know if that's possible

patriciareinoso (changes requested):

proto/server/subscriber.go
```diff
@@ -0,0 +1,148 @@
...
+		logger.DbLog.Debugf("insert/update authentication subscription in authenticationSubscription collection: %v", imsi)
+		filter := bson.M{"ueId": imsi}
+		authDataBsonA := configmodels.ToBsonM(authSubData)
+		authDataBsonA["ueId"] = imsi
+		_, err = dbadapter.AuthDBClient.RestfulAPIPost(authSubsDataColl, filter, authDataBsonA)
+		if err != nil {
+			if abortErr := session.AbortTransaction(sc); abortErr != nil {
+				logger.DbLog.Errorw("failed to abort transaction", "error", abortErr)
+			}
+		}
```
patriciareinoso, line 64: we need to return after transaction is aborted

proto/server/subscriber.go
```diff
@@ -0,0 +1,148 @@
...
+		logger.DbLog.Debugf("insert/update authentication subscription in amData collection: %v", imsi)
+		basicAmData := map[string]interface{}{
+			"ueId": imsi,
+		}
+		basicDataBson := configmodels.ToBsonM(basicAmData)
+		_, err = dbadapter.CommonDBClient.RestfulAPIPostWithContext(sc, amDataColl, filter, basicDataBson)
+		if err != nil {
+			if abortErr := session.AbortTransaction(sc); abortErr != nil {
+				logger.DbLog.Errorw("failed to abort transaction", "error", abortErr)
+			}
```
patriciareinoso, line 74: we need to return after transaction is aborted

patriciareinoso (changes requested):

proto/server/configEvtHandler.go
patriciareinoso: what about this comment?

proto/server/subscriber.go
```diff
@@ -0,0 +1,148 @@
...
+	defer session.EndSession(context.TODO())
+	err = mongo.WithSession(context.TODO(), session, func(sc mongo.SessionContext) error {
+		if err = session.StartTransaction(); err != nil {
+			return fmt.Errorf("failed to start transaction: %w", err)
+		}
+		logger.DbLog.Debugf("insert/update authentication subscription in authenticationSubscription collection: %v", imsi)
+		filter := bson.M{"ueId": imsi}
+		authDataBsonA := configmodels.ToBsonM(authSubData)
+		authDataBsonA["ueId"] = imsi
+		if _, err = dbadapter.AuthDBClient.RestfulAPIPost(authSubsDataColl, filter, authDataBsonA); err != nil {
```
patriciareinoso, line 59:
```suggestion
		if _, err = dbadapter.AuthDBClient.RestfulAPIPostWithContext(sc, authSubsDataColl, filter, authDataBsonA); err != nil {
```

patriciareinoso (changes requested):
when deleting a subscriber, the information in `amDataColl` is removed but the information in `authSubsDataColl` is not. 

I am getting the following error:
```
2025-03-07T10:30:04.169+0100	ERROR	server/subscriber.go:113	failed to delete subscriber imsi-208930100007487 from database: %!w(*errors.errorString=&{RestfulAPIDeleteOne err: session was not created by this client})	{"component": "WebUI", "category": "DB"}
```

In the POST operation, the writing operation in AuthDBClient is not using RestfulAPIPostWithContext so the transaction is not working well either. 

the error is making me think that maybe since we are using 2 different DBs this is not the way of using transaction or maybe they do not make sense at all. 

But if they do, please, test that the transactions work well.

proto/server/configEvtHandler.go
```diff
@@ -39,9 +39,10 @@ type Update5GSubscriberMsg struct {
...
-	execCommand = exec.Command
-	imsiData    map[string]*models.AuthenticationSubscription
-	rwLock      sync.RWMutex
+	execCommand        = exec.Command
+	imsiData           map[string]*models.AuthenticationSubscription
+	rwLock             sync.RWMutex
+	subscriberAuthData SubscriberAuthenticationData
 )
 
 func init() {
```
patriciareinoso, line 48: i don't think this is the place to initialized the imsiData anymore. it should be initialized together with subscriberAuthData
and moved from this file
https://github.com/omec-project/webconsole/pull/310/files#r1979782615

proto/server/subscriber.go
```diff
@@ -0,0 +1,148 @@
...
+
+type MemorySubscriberAuthenticationData struct {
+	SubscriberAuthenticationData
+}
+
+func (subscriberAuthData MemorySubscriberAuthenticationData) SubscriberAuthenticationDataGet(imsi string) (authSubData *models.AuthenticationSubscription) {
+	return imsiData[imsi]
+}
+
+func (subscriberAuthData MemorySubscriberAuthenticationData) SubscriberAuthenticationDataCreate(imsi string, authSubData *models.AuthenticationSubscription) {
```
patriciareinoso, line 124: we already know that the DB is not initialized for 4G. I don't think that it's worth to keep trying to write in the DB. WDYT?

proto/server/configEvtHandler.go
patriciareinoso, line 48: i don't think this is the place to initialized the imsiData anymore. it should be initialized together with subscriberAuthData
and moved from this file
https://github.com/omec-project/webconsole/pull/310/files#r1979782615

patriciareinoso, line 48: i don't think this init function is the place to initialize the imsiData anymore. it should be initialized together with subscriberAuthData
and moved from this file
https://github.com/omec-project/webconsole/pull/310/files#r1979782615

proto/server/subscriber.go
patriciareinoso: We saw that the transaction approach do not work in this case because we are using different DB clients

proto/server/subscriber.go
```diff
@@ -0,0 +1,110 @@
...
+		logger.DbLog.Warnln(err)
+	}
+}
+
+func (subscriberAuthData MemorySubscriberAuthenticationData) SubscriberAuthenticationDataDelete(imsi string) {
+	logger.WebUILog.Debugf("delete authentication subscription from amData collection: %v", imsi)
+	filter := bson.M{"ueId": imsi}
+	err := dbadapter.CommonDBClient.RestfulAPIDeleteOne(amDataColl, filter)
+	if err != nil {
+		logger.DbLog.Warnln(err)
```
gatici, line 105:
```suggestion
		logger.DbLog.Errorln(err)
```
Can we use  error log level to make it more visible ? 
This applies also the other lines which logs errors.

gatici:
I found a solution called Two-Phase Commit (2PC) which is a distributed algorithm used to ensure atomicity in a distributed transaction such as 2 different databases. Atomicity means that a transaction either completes successfully (commits) across all participants or fails entirely (rolls back), leaving the system in a consistent state. https://vinayaksb.wordpress.com/2015/05/08/distributed-transaction-with-nosql-using-2-phase-commit/ 
But I am not sure if it is worth to implement.

patriciareinoso (approved):

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+		PermanentKey: &models.PermanentKey{
+			EncryptionAlgorithm: 0,
+			EncryptionKey:       0,
+			PermanentKeyValue:   "8baf473f2f8fd09487cccbd7097c6862",
+		},
+		SequenceNumber: "16f3b3f70fc2",
+	}
+	subscribers := []bson.M{configmodels.ToBsonM(subscriber)}
+	subscribers[0]["ueId"] = "imsi-208930100007487"
+	dbadapter.AuthDBClient = &MockMongoSubscriberGetOne{dbadapter.AuthDBClient, subscribers[0]}
```
gatici, line 486: This modifies the global `dbadapter.AuthDBClient`. We need to set it to its original state using:
originalClient := dbadapter.AuthDBClient
defer func() { dbadapter.AuthDBClient = originalClient }()

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+	subscriberResult := subscriberAuthData.SubscriberAuthenticationDataGet("imsi-208930100007487")
+	if !reflect.DeepEqual(&subscriber, subscriberResult) {
+		t.Errorf("Expected subscriber %v, got %v", &subscriber, subscriberResult)
+	}
+}
+
+func Test_handleSubscriberGet4G(t *testing.T) {
+	subscriberAuthData = MemorySubscriberAuthenticationData{}
+	imsiData = make(map[string]*models.AuthenticationSubscription, 0)
```
gatici, line 495:
```suggestion
	imsiData = make(map[string]*models.AuthenticationSubscription)
```
Map does not require explicit size.

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+	subscribers[0]["ueId"] = "imsi-208930100007487"
+	dbadapter.AuthDBClient = &MockMongoSubscriberGetOne{dbadapter.AuthDBClient, subscribers[0]}
+	subscriberResult := subscriberAuthData.SubscriberAuthenticationDataGet("imsi-208930100007487")
+	if !reflect.DeepEqual(&subscriber, subscriberResult) {
+		t.Errorf("Expected subscriber %v, got %v", &subscriber, subscriberResult)
+	}
+}
+
+func Test_handleSubscriberGet4G(t *testing.T) {
+	subscriberAuthData = MemorySubscriberAuthenticationData{}
```
gatici, line 494: The test directly modifies global variables `subscriberAuthData` and `imsiData`. We need to restore them to their original states.

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+				EncryptionAlgorithm: 0,
+				EncryptionKey:       0,
+				PermanentKeyValue:   "8baf473f2f8fd09487cccbd7097c6862",
+			},
+			SequenceNumber: "16f3b3f70fc2",
+		},
 	}
 
 	postData = make([]map[string]interface{}, 0)
+	imsiData = make(map[string]*models.AuthenticationSubscription, 0)
```
gatici, line 552:
```suggestion
	imsiData = make(map[string]*models.AuthenticationSubscription)
```

gatici: The map does not require explicit size.

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+	subscriberResult := subscriberAuthData.SubscriberAuthenticationDataGet("imsi-208930100007487")
+	if !reflect.DeepEqual(&subscriber, subscriberResult) {
+		t.Errorf("Expected subscriber %v, got %v", &subscriber, subscriberResult)
+	}
+}
+
+func Test_handleSubscriberPost5G(t *testing.T) {
 	ueId := "208930100007487"
-	factory.WebUIConfig.Configuration.Mode5G = true
+	subscriberAuthData = DatabaseSubscriberAuthenticationData{}
```
gatici, line 526: This test modifies global subscriberAuthData, imsiData, dbadapter.AuthDBClient, dbadapter.CommonDBClient. We need to restore them to their original states.

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+				EncryptionAlgorithm: 0,
+				EncryptionKey:       0,
+				PermanentKeyValue:   "8baf473f2f8fd09487cccbd7097c6862",
+			},
+			SequenceNumber: "16f3b3f70fc2",
+		},
+	}
+
+	postData = make([]map[string]interface{}, 0)
+	imsiData = make(map[string]*models.AuthenticationSubscription, 0)
```
gatici, line 620:
```suggestion
	imsiData = make(map[string]*models.AuthenticationSubscription)
```

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+	dbadapter.CommonDBClient = &MockMongoPost{}
```
gatici, line 621: dbadapter.CommonDBClient and subscriberAuthData  instances depend on global variables and we need to restore them back.

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
-	var result map[string]interface{} = postData[0]["data"].(map[string]interface{})
-	if result["ueId"] != ueId {
-		t.Errorf("Expected ueId %v, got %v", ueId, result["ueId"])
+	var AmDataResult map[string]interface{} = postData[0]["data"].(map[string]interface{})
+	if AmDataResult["ueId"] != ueId {
+		t.Errorf("Expected ueId %v, got %v", ueId, AmDataResult["ueId"])
+	}
+	if imsiData[ueId] == nil {
+		t.Errorf("Expected ueId %v in memory, got nil", ueId)
+	}
```
gatici, line 640: This can be removed as the next lines validate it.

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+		t.Errorf("Expected authSubData %v in memory, got %v ", configMsg.AuthSubData, imsiData[ueId])
+	}
+}
+
+func Test_handleSubscriberDelete5G(t *testing.T) {
+	ueId := "208930100007487"
+	subscriberAuthData = DatabaseSubscriberAuthenticationData{}
+
+	deleteData = make([]map[string]interface{}, 0)
+	dbadapter.AuthDBClient = &MockMongoDeleteOne{}
```
gatici, line 651: The function changes global variables `subscriberAuthData`, `dbadapter.AuthDBClient`, `dbadapter.CommonDBClient`, `deleteData` but it does not restore them back.

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+		t.Errorf("Expected filter %v, got %v", expectedFilter, deleteData[1]["filter"])
+	}
+}
+
+func Test_handleSubscriberDelete4G(t *testing.T) {
+	ueId := "208930100007487"
+	subscriberAuthData = MemorySubscriberAuthenticationData{}
+
+	deleteData = make([]map[string]interface{}, 0)
+	imsiData = make(map[string]*models.AuthenticationSubscription, 0)
```
gatici, line 678:
```suggestion
	imsiData = make(map[string]*models.AuthenticationSubscription)
```

gatici: This is not wrong but unnecessary.

proto/server/configEvtHandler_test.go
```diff
@@ -439,31 +458,260 @@ func Test_handleNetworkSlicePost_alreadyExists(t *testing.T) {
...
+	if !reflect.DeepEqual(deleteData[1]["filter"], expectedFilter) {
+		t.Errorf("Expected filter %v, got %v", expectedFilter, deleteData[1]["filter"])
+	}
+}
+
+func Test_handleSubscriberDelete4G(t *testing.T) {
+	ueId := "208930100007487"
+	subscriberAuthData = MemorySubscriberAuthenticationData{}
+
+	deleteData = make([]map[string]interface{}, 0)
```
gatici, line 677: This function modifies global variables `deleteData`, `imsiData`, `subscriberAuthData`,
`dbadapter.CommonDBClient`. We need set them back to their original states.

gatici (commented):
Looks good. I gave some minor comments regarding tests.

ghislainbourgeois (approved):
LGTM

PR #314, opened 2025-02-27
fix: improve regex used for name and FQDN validation

patriciareinoso:
This PR improve the regex used for name validation and FQDN validate:

1. Names only starts with letters
2. FQDN can start with letters or numbers but not `-`


gab-arrobo (approved):
It looks good to me.... just one question, any reason why the name cannot start with a number?
FYI, ROC does this kind of validation for the device groups (their name cannot start with a number)

Issue #301, opened 2025-02-05
Delete Device Group / Network Slice also deletes the Subscribers from DB


patriciareinoso:
## Description

When we create a subscriber X (POST), the following tables are created: 
- `subscriptionData.authenticationData.authenticationSubscription`
- `subscriptionData.provisionedData.amData`

When subscriber X is associated to a DG that belongs to a NS, the following tables are created:
- `subscriptionData.provisionedData.smData`
- `subscriptionData.provisionedData.smfSelectionSubscriptionData`
- `policyData.ues.amData`
- `policyData.ues.smData`
- `policyData.ues.flowRule`
- AND `subscriptionData.provisionedData.amData` is updated to add PlmnID, Nssai, Gpsis, SubscribedUeAmbr

When we DELETE a DG or a NS, the following tables are deleted:
- `subscriptionData.provisionedData.smData`
- `subscriptionData.provisionedData.smfSelectionSubscriptionData`
- `policyData.ues.amData`
- `policyData.ues.smData`
- `policyData.ues.flowRule`
- ### **INCLUDING  `subscriptionData.provisionedData.amData`**

## The Issue

When we retrieve all subcribers `GET /api/subscriber`, the information returned is based on `subscriptionData.provisionedData.amData` table. Which mean that if we try to retrieve all the subscribers, the subscriber X will no longer be on the list. And we'll have "orphan" information in the `subscriptionData.authenticationData.authenticationSubscription` table.

If we retrieve the subscriber by name (`GET /api/subscriber/X`) you'll  see that the `authenticationSubscription` is filled but the rest of the fields are empty.

## Expected behavior

I would have expected that when I delete a DG/NS, the subscribers associated to it are not removed. The same way a DG is not removed when you delete a NS that contains it.

So I would expect that `GET /api/subscriber` contains this subscriber

## How to reproduce

1. Create a Subcriber
2. Create a DG and associate the subscriber to it
3. Create a NS and associate the the DG to it.

Check the DB tables.

4. Delete the DG.

Only the `subscriptionData.authenticationData.authenticationSubscription` should exist. 
And when try to   `GET /api/subscriber` the subscriber does not exist anymore.
If you `GET /api/subscriber/X` you'll  see that the  `authenticationSubscription` is filled but the rest of the fields are empty.

## Solution attempts

[...]

PR #317, opened 2025-03-06
feat: return 409 conflict error if a subscriber already exists

gatici:
Description:

This PR aims to provide uniqueness for a subscriber's IMSI in the database. 
It protects to override an existing subscriber by returning  `HTTP 409 (Conflict)` error If a subscriber with the given IMSI already exists in the Database.

Previous Behavior:

It does not check if the subscriber with same IMSI exists or not. If the POST request send again, it will override the existing subscriber data.

After this PR:

If a subscriber with the same IMSI exists, it return 409 Conflict error and the existing subscriber data is protected.

### DIFF
--- a/configapi/api_sub_config.go
+++ b/configapi/api_sub_config.go
@@ -462,6 +462,18 @@ func PostSubscriberByID(c *gin.Context) {
 
 	logger.WebUILog.Infoln("Received Post Subscriber Data from Roc/Simapp: ", ueId)
 
+	// Check if the IMSI already exists in the database
+	filter := bson.M{"ueId": ueId}
+	subscriber, err := dbadapter.CommonDBClient.RestfulAPIGetOne(amDataColl, filter)
+	if err != nil {
+		logger.DbLog.Errorw("failed to check subscriber existence", "IMSI", ueId, "error", err)
+		c.JSON(http.StatusInternalServerError, gin.H{"error": fmt.Sprintf("failed to check subscriber existence: %v", err)})
+		return
+	} else if subscriber != nil {
+		logger.WebUILog.Errorf("subscriber with IMSI %s already exists", ueId)
+		c.JSON(http.StatusConflict, gin.H{"error": fmt.Sprintf("subscriber with IMSI %s already exists", ueId)})
+		return
+	}
 	authSubsData := models.AuthenticationSubscription{
 		AuthenticationManagementField: "8000",
 		AuthenticationMethod:          "5G_AKA", // "5G_AKA", "EAP_AKA_PRIME"
--- a/configapi/api_sub_config.go
+++ b/configapi/api_sub_config.go
@@ -446,6 +446,7 @@ func GetSubscriberByID(c *gin.Context) {
 // @Failure      400  {object}  nil  "Invalid subscriber content"
 // @Failure      401  {object}  nil  "Authorization failed"
 // @Failure      403  {object}  nil  "Forbidden"
+// @Failure      409  {object}  nil  "Subscriber already exists"
 // @Failure      500  {object}  nil  "Error creating subscriber"
 // @Router      /api/subscriber/{imsi}  [post]
 func PostSubscriberByID(c *gin.Context) {
@@ -466,12 +467,12 @@ func PostSubscriberByID(c *gin.Context) {
 	filter := bson.M{"ueId": ueId}
 	subscriber, err := dbadapter.CommonDBClient.RestfulAPIGetOne(amDataColl, filter)
 	if err != nil {
-		logger.DbLog.Errorw("failed to check subscriber existence", "IMSI", ueId, "error", err)
-		c.JSON(http.StatusInternalServerError, gin.H{"error": fmt.Sprintf("failed to check subscriber existence: %v", err)})
+		logger.DbLog.Errorf("Failed querying subscriber existence for IMSI: %s; Error: %v", ueId, err)
+		c.JSON(http.StatusInternalServerError, gin.H{"error": fmt.Sprintf("failed to check subscriber: %s existence", ueId)})
 		return
 	} else if subscriber != nil {
-		logger.WebUILog.Errorf("subscriber with IMSI %s already exists", ueId)
-		c.JSON(http.StatusConflict, gin.H{"error": fmt.Sprintf("subscriber with IMSI %s already exists", ueId)})
+		logger.WebUILog.Errorf("subscriber %s already exists", ueId)
+		c.JSON(http.StatusConflict, gin.H{"error": fmt.Sprintf("subscriber %s already exists", ueId)})
 		return
 	}
 	authSubsData := models.AuthenticationSubscription{
--- a/configapi/api_sub_config.go
+++ b/configapi/api_sub_config.go
@@ -467,7 +467,7 @@ func PostSubscriberByID(c *gin.Context) {
 	filter := bson.M{"ueId": ueId}
 	subscriber, err := dbadapter.CommonDBClient.RestfulAPIGetOne(amDataColl, filter)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


configapi/api_sub_config_test.go
```diff
@@ -224,6 +240,73 @@ func TestSubscriberPostHandlers(t *testing.T) {
...
+		{
+			name:            "Create subscriber conflict subscriber already exists",
+			route:           "/api/subscriber/imsi-208930100007487",
+			inputData:       `{"plmnID":"12345", "opc":"8e27b6af0e692e750f32667a3b14605d", "key":"8baf473f2f8fd09487cccbd7097c6862", "sequenceNumber":"16f3b3f70fc2"}`,
+			expectedCode:    http.StatusConflict,
+			expectedBody:    `{"error":"subscriber already exists"}`,
+			expectedMessage: nil,
+		},
+	}
+	for _, tc := range testCases {
```
Gmerold, line 265: Do we need this `for`? There's only one test case anyway.

configapi/api_sub_config.go
```diff
@@ -392,6 +392,18 @@ func PostSubscriberByID(c *gin.Context) {
```
patriciareinoso, line 392: Please add this line to the docstrings:
```
// @Failure 409 {object} nil "Subscriber already exists"
```
(or something similar)
and then run:
```
swag init -g backend/webui_service/swagger_ui_service.go --outputTypes go
```
to generate the new swagger UI template

configapi/api_sub_config.go
```diff
@@ -392,6 +392,18 @@ func PostSubscriberByID(c *gin.Context) {
 
 	logger.WebUILog.Infoln("Received Post Subscriber Data from Roc/Simapp: ", ueId)
 
+	// Check if the IMSI already exists in the database
+	filter := bson.M{"ueId": ueId}
+	subscriber, err := dbadapter.CommonDBClient.RestfulAPIGetOne(amDataColl, filter)
+	if err != nil {
+		logger.DbLog.Errorw("failed to check subscriber existence", "IMSI", ueId, "error", err)
+		c.JSON(http.StatusInternalServerError, gin.H{"error": fmt.Sprintf("failed to check subscriber existence: %v", err)})
```
patriciareinoso, line 400: i would not include too many details in the user response

configapi/api_sub_config.go
```diff
@@ -392,6 +392,18 @@ func PostSubscriberByID(c *gin.Context) {
...
+		return
+	} else if subscriber != nil {
+		logger.WebUILog.Errorf("subscriber with IMSI %s already exists", ueId)
+		c.JSON(http.StatusConflict, gin.H{"error": fmt.Sprintf("subscriber with IMSI %s already exists", ueId)})
```
patriciareinoso, line 404:
```suggestion
		logger.WebUILog.Errorf("subscriber %s already exists", ueId)
		c.JSON(http.StatusConflict, gin.H{"error": fmt.Sprintf("subscriber %s already exists", ueId)})
```
the ueId is `imsi-XXXX` so it seems redundant, but this is just personal preference

configapi/api_sub_config_test.go
```diff
@@ -145,13 +155,17 @@ func TestSubscriberPostHandlers(t *testing.T) {
 		name            string
 		route           string
 		inputData       string
-		expectedMessage configmodels.ConfigMessage
+		expectedCode    int
+		expectedBody    string
```
patriciareinoso, line 159:
```suggestion

```
expectedCode and expectedBody are overriden during test

configapi/api_sub_config_test.go
```diff
@@ -224,6 +240,73 @@ func TestSubscriberPostHandlers(t *testing.T) {
...
+				default:
+					// No message received, as expected
+				}
+			} else {
+				select {
+				case <-configChannel:
+				default:
+					t.Error("expected message in configChannel, but none received")
+				}
+			}
```
patriciareinoso, line 305:
```suggestion
			select {
			case <-configChannel:
				t.Error("expected no message in configChannel, but got one")
			default:
				// No message received, as expected
			}

```

configapi/api_sub_config_test.go
```diff
@@ -224,6 +240,73 @@ func TestSubscriberPostHandlers(t *testing.T) {
...
+	router := gin.Default()
+	AddApiService(router)
+
+	testCases := []struct {
+		name            string
+		route           string
+		inputData       string
+		expectedCode    int
+		expectedBody    string
+		expectedMessage *configmodels.ConfigMessage
```
patriciareinoso, line 254:
```suggestion

```
expectedCode and expectedBody are overriden during test
and expectedMessage is not really needed

configapi/api_sub_config.go
```diff
@@ -392,6 +392,18 @@ func PostSubscriberByID(c *gin.Context) {
 
 	logger.WebUILog.Infoln("Received Post Subscriber Data from Roc/Simapp: ", ueId)
 
+	// Check if the IMSI already exists in the database
+	filter := bson.M{"ueId": ueId}
+	subscriber, err := dbadapter.CommonDBClient.RestfulAPIGetOne(amDataColl, filter)
```
patriciareinoso, line 397: a comment about this DB retrieval:

There is an issue (https://github.com/omec-project/webconsole/issues/301) in which if we remove the subscriber imsi-XXX from the list of subscribers in a DG, and we do not do a explicit DELETE /api/subscriber/imsi-XXX, the entry in `amDataColl` for subscriber imsi-XXX will be removed but the entry in `authSubsDataColl` will not be removed.
So there is a potential inconsistency here : 
1. If we list the subscribers (GET /api/subscriber) -> we will not see the subscriber imsi-XXX (because it's based on `amDataColl`)
2. In here we will consider that it does not exists
3. But After your modifications in GET /api/subscirber/imsi-XXX, it will retrieve a subscriber only with auth information because the information in `authSubsDataColl` would still exist.

My intention with this comment is not for you to fix this issue, it's just to create awareness.  (and I think that the problem is in the delete operation, not in the getter).
In the NMS we have protection for this case.

patriciareinoso (changes requested):

configapi/api_sub_config.go
gatici: The error details are removed from the response.

configapi/api_sub_config.go
gatici: Done.

configapi/api_sub_config.go
gatici: OK, removed.

configapi/api_sub_config_test.go
gatici: No need. Removed.

configapi/api_sub_config_test.go
gatici: Test case struct is removed.

configapi/api_sub_config_test.go
gatici: Done

configapi/api_sub_config_test.go
gatici: Test case struct definition is removed.

configapi/api_sub_config.go
gatici: OK, We can try to fix it in a different PR. I need to think about it.

patriciareinoso (approved):

patriciareinoso (commented):
I have been thinking that we need to add the go doc for the PUT operation so that we can use it in the swagger UI.

gatici:
> I have been thinking that we need to add the go doc for the PUT operation so that we can use it in the swagger UI.

The go doc is added for PUT operation.

gab-arrobo (approved):
+1

Issue #109, opened 2023-12-13
GET a subscriber that does not exist returns an object


patriciareinoso:
### Describe the bug

If we try to GET a subscriber that does not exists, the result is some kind of empty subscriber object with `ueId` equal to the imsi we specified

#### To Reproduce

`curl -vvv ${WEBUI_IP}:5000/api/subscriber/<subscriber>` 

where `<subscriber>` does not exist. 

The output will be :

`{"plmnID":"","ueId":<subscriber>,"AuthenticationSubscription":{"authenticationMethod":"","permanentKey":null,"sequenceNumber":""},"AccessAndMobilitySubscriptionData":{},"SessionManagementSubscriptionData":null,"SmfSelectionSubscriptionData":{},"AmPolicyData":{},"SmPolicyData":{"smPolicySnssaiData":null},"FlowRules":null}`

#### Expected behavior

Subscriber not found. 404 error. 

#### Screenshots

![image](https://github.com/omec-project/webconsole/assets/10354025/c9aa7cf6-9e7d-47c1-946f-29c597a62436)

PR #318, opened 2025-03-06
feat: return 404 if a subscriber does not exist

gatici:
Previous Behavior:
If a subscriber is not found in the DB, it returns an object with empty values as below:

```
{"plmnID":"","ueId":<subscriber>,"AuthenticationSubscription":{"authenticationMethod":"","permanentKey":null,"sequenceNumber":""},"AccessAndMobilitySubscriptionData":{},"SessionManagementSubscriptionData":null,"SmfSelectionSubscriptionData":{},"AmPolicyData":{},"SmPolicyData":{"smPolicySnssaiData":null},"FlowRules":null}
```

After this PR:

If a subscriber is not found in the DB, it will respond with `404 Not Found`.

Fixes https://github.com/omec-project/webconsole/issues/109.

### DIFF
--- a/configapi/api_sub_config.go
+++ b/configapi/api_sub_config.go
@@ -316,37 +316,37 @@ func GetSubscriberByID(c *gin.Context) {
 
 	authSubsDataInterface, err := dbadapter.AuthDBClient.RestfulAPIGetOne(authSubsDataColl, filterUeIdOnly)
 	if err != nil {
-		logger.DbLog.Errorf("Failed to fetch authentication subscription data from DB: %v", err)
+		logger.DbLog.Errorf("failed to fetch authentication subscription data from DB: %v", err)
 		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to fetch the requested user record from DB"})
 		return
 	}
 	amDataDataInterface, err := dbadapter.CommonDBClient.RestfulAPIGetOne(amDataColl, filterUeIdOnly)
 	if err != nil {
-		logger.DbLog.Errorf("Failed to fetch am data from DB: %v", err)
+		logger.DbLog.Errorf("failed to fetch am data from DB: %v", err)
 		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to fetch the requested user record from DB"})
 		return
 	}
 	smDataDataInterface, err := dbadapter.CommonDBClient.RestfulAPIGetMany(smDataColl, filterUeIdOnly)
 	if err != nil {
-		logger.DbLog.Errorf("Failed to fetch sm data from DB: %v", err)
+		logger.DbLog.Errorf("failed to fetch sm data from DB: %v", err)
 		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to fetch the requested user record from DB"})
 		return
 	}
 	smfSelDataInterface, err := dbadapter.CommonDBClient.RestfulAPIGetOne(smfSelDataColl, filterUeIdOnly)
 	if err != nil {
-		logger.DbLog.Errorf("Failed to fetch smf selection data from DB: %v", err)
+		logger.DbLog.Errorf("failed to fetch smf selection data from DB: %v", err)
 		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to fetch the requested user record from DB"})
 		return
 	}
 	amPolicyDataInterface, err := dbadapter.CommonDBClient.RestfulAPIGetOne(amPolicyDataColl, filterUeIdOnly)
 	if err != nil {
-		logger.DbLog.Errorf("Failed to fetch am polict data from DB: %v", err)
+		logger.DbLog.Errorf("failed to fetch am policy data from DB: %v", err)
 		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to fetch the requested user record from DB"})
 		return
 	}
 	smPolicyDataInterface, err := dbadapter.CommonDBClient.RestfulAPIGetOne(smPolicyDataColl, filterUeIdOnly)
 	if err != nil {
-		logger.DbLog.Errorf("Failed to fetch sm policy data from DB: %v", err)
+		logger.DbLog.Errorf("failed to fetch sm policy data from DB: %v", err)
 		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to fetch the requested user record from DB"})
 		return
 	}
@@ -366,7 +366,7 @@ func GetSubscriberByID(c *gin.Context) {
 	if authSubsDataInterface != nil {
 		err := json.Unmarshal(configmodels.MapToByte(authSubsDataInterface), &authSubsData)
 		if err != nil {
-			logger.WebUILog.Errorf("Error unmarshalling authentication subscription data: %v", err)
+			logger.WebUILog.Errorf("error unmarshalling authentication subscription data: %v", err)
 			c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to retrieve user"})
 			return
 		}
@@ -376,7 +376,7 @@ func GetSubscriberByID(c *gin.Context) {
 	if amDataDataInterface != nil {
 		err := json.Unmarshal(configmodels.MapToByte(amDataDataInterface), &amDataData)
 		if err != nil {
-			logger.WebUILog.Errorf("Error unmarshalling access and mobility subscription data: %v", err)
+			logger.WebUILog.Errorf("error unmarshalling access and mobility subscription data: %v", err)
 			c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to retrieve user"})
 			return
 		}
@@ -386,7 +386,7 @@ func GetSubscriberByID(c *gin.Context) {
 	if smDataDataInterface != nil {
 		err := json.Unmarshal(sliceToByte(smDataDataInterface), &smDataData)
 		if err != nil {
-			logger.WebUILog.Errorf("Error unmarshalling session management subscription data: %v", err)
+			logger.WebUILog.Errorf("error unmarshalling session management subscription data: %v", err)
 			c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to retrieve user"})
 			return
 		}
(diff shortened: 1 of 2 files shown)


configapi/api_sub_config_test.go
```diff
@@ -5,6 +5,7 @@ package configapi
 
 import (
 	"encoding/json"
```
patriciareinoso, line 7: a general comment about the tests, i think that i see why this type of implementation for the mocks:
```
	if coll == "smDataColl" {
		return []map[string]interface{}{}, nil
	}
	return nil, fmt.Errorf("collection %s not found", coll)
```
however if the mock is called with any collection other than smDataColl, the test will not fail. It will only show some logs that we will probably not see. 
in `webconsole/proto/server/configEvtHandler_test.go`  there are examples in which we store the value used to call the mock and we can use that to do the checks. I suggest using this approach: we would be able to check that the collection was actually called with the expected value.

configapi/api_sub_config_test.go
```diff
@@ -72,6 +73,153 @@ func (m *MockMongoClientDeviceGroupsWithSubscriber) RestfulAPIGetMany(coll strin
...
+			expectedResponseContains: `"error":"subscriber with ID 12345 not found"`,
+		},
+		{
+			name:                     "Valid subscriber data retrieved",
+			ueId:                     "12345",
+			commonDbAdapter:          &MockCommonDBClientWithData{},
+			authDbAdapter:            &MockAuthDBClientWithData{},
+			route:                    "/api/subscriber/:ueId",
+			expectedHTTPStatus:       http.StatusOK,
+			expectedResponseContains: `"ueId":"12345"`,
```
patriciareinoso, line 193: can we check the whole content of the response?

configapi/api_sub_config.go
```diff
@@ -309,48 +309,100 @@ func GetSubscriberByID(c *gin.Context) {
...
+		c.JSON(http.StatusNotFound, gin.H{"error": fmt.Sprintf("subscriber with ID %s not found", ueId)})
+		return
 	}
+
 	var authSubsData models.AuthenticationSubscription
-	json.Unmarshal(configmodels.MapToByte(authSubsDataInterface), &authSubsData)
+	if authSubsDataInterface != nil {
+		err := json.Unmarshal(configmodels.MapToByte(authSubsDataInterface), &authSubsData)
+		if err != nil {
+			c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to unmarshal authentication subscription data"})
```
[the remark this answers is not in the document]
gatici: OK, The errors are logged with details but a generic error message is returned.

configapi/api_sub_config.go
```diff
@@ -309,48 +309,100 @@ func GetSubscriberByID(c *gin.Context) {
...
-
 	filterUeIdOnly := bson.M{"ueId": ueId}
 
-	authSubsDataInterface, errGetOneAuth := dbadapter.AuthDBClient.RestfulAPIGetOne(authSubsDataColl, filterUeIdOnly)
-	if errGetOneAuth != nil {
-		logger.DbLog.Warnln(errGetOneAuth)
+	var subsData configmodels.SubsData
+
+	authSubsDataInterface, err := dbadapter.AuthDBClient.RestfulAPIGetOne(authSubsDataColl, filterUeIdOnly)
+	if err != nil {
```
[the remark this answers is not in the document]
gatici: I changed it to return if an error occurs.

configapi/api_sub_config_test.go
```diff
@@ -72,6 +73,153 @@ func (m *MockMongoClientDeviceGroupsWithSubscriber) RestfulAPIGetMany(coll strin
 	return results, nil
 }
 
+type MockAuthDBClientEmpty struct {
+	dbadapter.DBInterface
+}
+
+func (m *MockAuthDBClientEmpty) RestfulAPIGetOne(coll string, filter bson.M) (map[string]interface{}, error) {
+	if coll == "authSubsDataColl" {
+		return nil, fmt.Errorf("no data found in collection %s", coll)
```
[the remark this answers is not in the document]
gatici: Done

configapi/api_sub_config_test.go
```diff
@@ -72,6 +73,153 @@ func (m *MockMongoClientDeviceGroupsWithSubscriber) RestfulAPIGetMany(coll strin
...
+}
+
+type MockCommonDBClientEmpty struct {
+	dbadapter.DBInterface
+}
+
+func (m *MockCommonDBClientEmpty) RestfulAPIGetOne(coll string, filter bson.M) (map[string]interface{}, error) {
+	switch coll {
+	case "amDataColl", "smfSelDataColl", "amPolicyDataColl", "smPolicyDataColl":
+		return nil, fmt.Errorf("no data found in collection %s", coll)
```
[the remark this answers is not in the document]
gatici: Done

configapi/api_sub_config_test.go
gatici: The postData is used to track the calls made to mock adapter methods during tests. It  records the queries to collections then we can validate the collections that a query is executed.

configapi/api_sub_config_test.go
gatici: The whole response is checked.

configapi/api_sub_config_test.go
```diff
@@ -72,6 +73,153 @@ func (m *MockMongoClientDeviceGroupsWithSubscriber) RestfulAPIGetMany(coll strin
...
+	default:
+		return nil, fmt.Errorf("collection %s not found", coll)
+	}
+}
+
+func (m *MockCommonDBClientEmpty) RestfulAPIGetMany(coll string, filter bson.M) ([]map[string]interface{}, error) {
+	if coll == "smDataColl" {
+		return []map[string]interface{}{}, nil
```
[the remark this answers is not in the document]
gatici: Fixed.

configapi/api_sub_config_test.go
```diff
@@ -72,6 +73,274 @@ func (m *MockMongoClientDeviceGroupsWithSubscriber) RestfulAPIGetMany(coll strin
...
+			{
+				"ueId": filter["ueId"],
+				"data": "session management data",
+			},
+		}, nil
+	}
+	return nil, fmt.Errorf("collection %s not found", coll)
+}
+
+func comparePostData(expected, actual []map[string]interface{}) error {
```
patriciareinoso, line 232: can we use `reflect.DeepEqual` to compare ?

configapi/api_sub_config_test.go
```diff
@@ -72,6 +73,274 @@ func (m *MockMongoClientDeviceGroupsWithSubscriber) RestfulAPIGetMany(coll strin
...
+		},
+
+		{
+			name:                 "Valid subscriber data retrieved",
+			ueId:                 "12345",
+			commonDbAdapter:      &MockCommonDBClientWithData{PostData: &postData},
+			authDbAdapter:        &MockAuthDBClientWithData{PostData: &postData},
+			route:                "/api/subscriber/:ueId",
+			expectedHTTPStatus:   http.StatusOK,
+			expectedFullResponse: `{"plmnID":"","ueId":"12345","AuthenticationSubscription":{"authenticationMethod":"5G_AKA","permanentKey":{"permanentKeyValue":"","encryptionKey":0,"encryptionAlgorithm":0},"sequenceNumber":"16f3b3f70fc2","authenticationManagementField":"8000","milenage":{"op":{"opValue":"","encryptionKey":0,"encryptionAlgorithm":0}},"opc":{"opcValue":"8e27b6af0e692e750f32667a3b14605d","encryptionKey":0,"encryptionAlgorithm":0}},"AccessAndMobilitySubscriptionData":{},"SessionManagementSubscriptionData":[{"singleNssai":null}],"SmfSelectionSubscriptionData":{"subscribedSnssaiInfos":{"010203":{"dnnInfos":[{"dnn":"internet"}]}}},"AmPolicyData":{},"SmPolicyData":{"smPolicySnssaiData":null},"FlowRules":null}`,
```
patriciareinoso, line 304: this line is too long. is there a way to cut into several lines?

patriciareinoso: I think there is a big mismatch between the information that is being returned by the mocks and the object returned in the response.

for example: the fields returned by the mock `(m *MockAuthDBClientWithData) RestfulAPIGetOne` in the case `policyData.ues.amData` (ueId and amData) do not correspond at all to the fields in AmPolicyData struct which is then, the one used to unmarshall. This results in an empty `AmPolicyData` struct in the response object. So the information returned by the mock is never used in any check.

I am not a fan of adding logic to the mock, but i understand that this is hard case because of the multiple calls. So i can think of 2 solutions:

1. We can return a generic object that can be unmarshaled by all the interfaces without failing and not containing much meaningful information. We do this in no matter the value of column used in the mock call. Then, assert that the returned object is equal to the empty structs in configmodels.SubsData. 
Like this we can get rid of the complicated logic in the return values in the mock.
We consider that checking that the mock is called with the proper column and filter is enough.

2. If we want to be more exhaustive in the checks, `(m *MockAuthDBClientWithData) RestfulAPIGetOne` can return a object containing a bunch of fields corresponding to each struct in configmodels.SubsData, that can be unmarshalled by all the interfaces. Like this, each struct will unmarshal only what is meaningful to them, and in the response object we'll have non-empty fields. We can remove the logic in the mocks.

We can discuss this if you want. It's quite complicated whay i am trying to say here.

configapi/api_sub_config_test.go
```diff
@@ -72,6 +73,274 @@ func (m *MockMongoClientDeviceGroupsWithSubscriber) RestfulAPIGetMany(coll strin
...
+}
+
+func (m *MockAuthDBClientWithData) RestfulAPIGetOne(coll string, filter bson.M) (map[string]interface{}, error) {
+	if m.PostData != nil {
+		*m.PostData = append(*m.PostData, map[string]interface{}{
+			"coll":   coll,
+			"filter": filter,
+		})
+	}
+	if coll == "subscriptionData.authenticationData.authenticationSubscription" && filter["ueId"] != nil {
```
patriciareinoso, line 103: i think we can remove this type of check and just return the value,
When checking the content of m.PosData we'll assert that it was called with the proper column name.
wdyt?

configapi/api_sub_config_test.go
gatici: `reflect.DeepEqual` is used to compare the expected responses.

configapi/api_sub_config_test.go
gatici: I followed the 2. approach. The DB mocks return objects corresponding to the structs from  configmodels.SubsData. I thought this is much more closet to a real testing.

patriciareinoso (approved):

gab-arrobo (approved):
+1

PR #328, opened 2025-03-14
feat: improve gNB validation in Network Slice creation

dariofaccin:
This PR aims to improve the validation of the gNBs included in the `site-info` object during the creation/editing of a Network Slice.
After this PR, a network slice cannot be created if any of the specified gNBs do not satisfy these requirements:
* `name` matches against the following regex: `^[a-zA-Z][a-zA-Z0-9-_]+$`
* `tac` is a valid integer in the following range: `[1, 16777215]`

This PR also improves the user experience by providing a meaningful error message in the HTTP response body when the Network Slice can not be created due to validation errors.

Type of `Tac` is changed from `string` to `*int32` for these reasons:
* same type across the whole webconsole
* pointers allow to distinguish between a nil value (which should be allowed, meaning no TAC is set in the request body) and a zero value (which should not be allowed)


configapi/api_slice_mgmt.go
```diff
@@ -168,11 +169,12 @@ func NetworkSliceDeleteHandler(c *gin.Context) bool {
...
 
-func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
+func NetworkSlicePostHandler(c *gin.Context, msgOp int) (bool, string) {
 	sliceName, _ := c.Params.Get("slice-name")
 	if !isValidName(sliceName) {
-		logger.ConfigLog.Errorf("invalid Network Slice name %s. Name needs to match the following regular expression: %s", sliceName, NAME_PATTERN)
-		return false
+		errorMsg := fmt.Sprintf("invalid Network Slice name %s. Name needs to match the following regular expression: %s", sliceName, NAME_PATTERN)
+		logger.ConfigLog.Errorf(errorMsg)
+		return false, errorMsg
```
ghislainbourgeois, line 177: Following my suggestion to change the signature of the function, this code would change in a way similar to this:

```suggestion
		err := fmt.Errorf("invalid Network Slice name %s. Name needs to match the following regular expression: %s", sliceName, NAME_PATTERN)
		logger.ConfigLog.Error(err.Error())
		return err
```

At the end of the function, we would then explicitly return nil.

ghislainbourgeois (changes requested):

configapi/api_default.go
```diff
@@ -242,19 +242,19 @@ func NetworkSliceSliceNameDelete(c *gin.Context) {
...
 
 // NetworkSliceSliceNamePut -
 func NetworkSliceSliceNamePut(c *gin.Context) {
 	logger.ConfigLog.Debugf("Received NetworkSliceSliceNamePut ")
-	if ret := NetworkSlicePostHandler(c, configmodels.Put_op); ret {
-		c.JSON(http.StatusOK, gin.H{})
-	} else {
-		c.JSON(http.StatusBadRequest, gin.H{})
+	if err := NetworkSlicePostHandler(c, configmodels.Put_op); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err})
```
ghislainbourgeois, line 256:
```suggestion
		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
```

configapi/api_slice_mgmt.go
```diff
@@ -184,8 +186,8 @@ func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
 		err = c.ShouldBindJSON(&request)
 	}
 	if err != nil {
-		logger.ConfigLog.Infof("err %v", err)
-		return false
+		logger.ConfigLog.Errorf("err %v", err)
```
ghislainbourgeois, line 189:
```suggestion
		logger.ConfigLog.Error(err.Error())
```

I would try to standardize the form that we use in this function. This seems sufficient.

configapi/api_slice_mgmt.go
```diff
@@ -198,6 +200,19 @@ func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
 	logger.ConfigLog.Infof("url: %v ", req.URL)
 	procReq := req.Body.(configmodels.Slice)
 
+	for _, gnb := range procReq.SiteInfo.GNodeBs {
+		if !isValidName(gnb.Name) {
+			err := fmt.Errorf("invalid gNB name `%s` in Network Slice %s. Name needs to match the following regular expression: %s", gnb.Name, sliceName, NAME_PATTERN)
+			logger.ConfigLog.Errorln(err.Error())
```
ghislainbourgeois, line 206:
```suggestion
			logger.ConfigLog.Error(err.Error())
```

Again to standardize the form.

ghislainbourgeois (changes requested):
I am pretty happy with this, just have a couple of tiny suggestions.

configapi/api_slice_mgmt_test.go
```diff
@@ -90,6 +91,23 @@ const DEVICE_GROUP_CONFIG = `{
   "site-info": "string"
 }`
 
+func networkSliceInvalidGnb(gnbName string, gnbTac int32) string {
```
thakurajayL, line 94: Could we change this name to something else. I think this function is creating Network Slice with whatever gNB information is passed. Why should it be called InvalidGnb?

configapi/api_slice_mgmt.go
```diff
@@ -198,14 +200,27 @@ func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
...
 
 	group := procReq.SiteDeviceGroup
 	slices.Sort(group)
-	slices.Compact(group)
+	group = slices.Compact(group)
 	logger.ConfigLog.Infof("number of device groups %v", len(group))
-	for i := 0; i < len(group); i++ {
+	for i := range group {
 		logger.ConfigLog.Infof("device groups(%v) - %v", i+1, group[i])
 	}
```
patriciareinoso, line 225:
```suggestion
	for i, g := range group {
		logger.ConfigLog.Infof("device groups(%v) - %v", i+1, g)
	}
```
since you are already changing to use range.

configapi/api_slice_mgmt.go
```diff
@@ -198,14 +200,27 @@ func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
...
 
+	for _, gnb := range procReq.SiteInfo.GNodeBs {
+		if !isValidName(gnb.Name) {
+			err := fmt.Errorf("invalid gNB name `%s` in Network Slice %s. Name needs to match the following regular expression: %s", gnb.Name, sliceName, NAME_PATTERN)
+			logger.ConfigLog.Error(err.Error())
+			return err
+		}
+		castedGnbTac := fmt.Sprint(gnb.Tac)
+		if !isValidGnbTac(castedGnbTac) {
+			err := fmt.Errorf("invalid TAC %s for gNB %s in Network Slice %s. TAC must be a numeric string within the range [1, 16777215]", castedGnbTac, gnb.Name, sliceName)
```
patriciareinoso, line 211:
```suggestion
			err := fmt.Errorf("invalid TAC %s for gNB %s in Network Slice %s. TAC must be an integer within the range [1, 16777215]", castedGnbTac, gnb.Name, sliceName)
```
in the context of a NS, the gNB is an int

thakurajayL (approved):

configapi/api_slice_mgmt.go
```diff
@@ -198,15 +200,28 @@ func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
 	logger.ConfigLog.Infof("url: %v ", req.URL)
 	procReq := req.Body.(configmodels.Slice)
 
+	for _, gnb := range procReq.SiteInfo.GNodeBs {
+		if !isValidName(gnb.Name) {
+			err := fmt.Errorf("invalid gNB name `%s` in Network Slice %s. Name needs to match the following regular expression: %s", gnb.Name, sliceName, NAME_PATTERN)
+			logger.ConfigLog.Error(err.Error())
+			return err
+		}
+		castedGnbTac := fmt.Sprint(gnb.Tac)
```
ghislainbourgeois, line 209:
```suggestion
		castedGnbTac := strconv.Itoa(gnb.Tac)
```

This would be a bit more efficient, but this is very minor.

ghislainbourgeois (dismissed):
LGTM, I left an optional suggestions. Please update your branch with the latest changes from main.

configapi/api_slice_mgmt.go
```diff
@@ -184,8 +186,8 @@ func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
 		err = c.ShouldBindJSON(&request)
 	}
 	if err != nil {
-		logger.ConfigLog.Infof("err %v", err)
-		return false
+		logger.ConfigLog.Error(err.Error())
```
gab-arrobo, line 189:
```suggestion
		logger.ConfigLog.Errorln(err.Error())
```

configapi/api_slice_mgmt.go
```diff
@@ -198,15 +200,28 @@ func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
 	logger.ConfigLog.Infof("url: %v ", req.URL)
 	procReq := req.Body.(configmodels.Slice)
 
+	for _, gnb := range procReq.SiteInfo.GNodeBs {
+		if !isValidName(gnb.Name) {
+			err := fmt.Errorf("invalid gNB name `%s` in Network Slice %s. Name needs to match the following regular expression: %s", gnb.Name, sliceName, NAME_PATTERN)
+			logger.ConfigLog.Error(err.Error())
```
gab-arrobo, line 206:
```suggestion
			logger.ConfigLog.Errorln(err.Error())
```

configapi/api_slice_mgmt.go
```diff
@@ -198,15 +200,28 @@ func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
...
+			return err
+		}
+		castedGnbTac := fmt.Sprint(gnb.Tac)
+		if !isValidGnbTac(castedGnbTac) {
+			err := fmt.Errorf("invalid TAC %s for gNB %s in Network Slice %s. TAC must be an integer within the range [1, 16777215]", castedGnbTac, gnb.Name, sliceName)
+			logger.ConfigLog.Error(err.Error())
```
gab-arrobo, line 212:
```suggestion
			logger.ConfigLog.Errorln(err.Error())
```

configapi/api_slice_mgmt.go
```diff
@@ -168,11 +169,12 @@ func NetworkSliceDeleteHandler(c *gin.Context) bool {
...
 }
 
-func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
+func NetworkSlicePostHandler(c *gin.Context, msgOp int) error {
 	sliceName, _ := c.Params.Get("slice-name")
 	if !isValidName(sliceName) {
-		logger.ConfigLog.Errorf("invalid Network Slice name %s. Name needs to match the following regular expression: %s", sliceName, NAME_PATTERN)
-		return false
+		err := fmt.Errorf("invalid Network Slice name %s. Name needs to match the following regular expression: %s", sliceName, NAME_PATTERN)
+		logger.ConfigLog.Error(err.Error())
```
gab-arrobo, line 176:
```suggestion
		logger.ConfigLog.Errorln(err.Error())
```

gab-arrobo (changes requested):

gab-arrobo:
@dariofaccin, it looks like the commits have not been address as of yet (it may be that you are missing a commit that addresses my requested changes)

dariofaccin:
@gab-arrobo this can be merged now.

configapi/api_slice_mgmt.go
gab-arrobo: Would not it be better to properly update `configmodels/model_inventory.go` to have `Tac` as a number as follows:
```diff
type PutGnbRequest struct {
-	Tac string `json:"tac"`
+	Tac int32 `json:"tac"`
}
```
Or, alternatively, to update `configmodels/model_slice_site_info_g_node_bs.go` to have `Tac` as string as follows:
```diff
type SliceSiteInfoGNodeBs struct {
	Name string `json:"name,omitempty"`

	// unique tac per gNB. This should match gNB configuration.
-	Tac int32 `json:"tac,omitempty"`
+	Tac string `json:"tac,omitempty"`
}
```
It looks pretty strange to me that we have `Tac` as string in one place while we have `Tac` as int32 in another place

configapi/api_slice_mgmt.go
gab-arrobo: It looks like this comment has not been addressed yet.

configapi/api_slice_mgmt.go
```diff
@@ -198,15 +200,28 @@ func NetworkSlicePostHandler(c *gin.Context, msgOp int) bool {
...
 	procReq := req.Body.(configmodels.Slice)
 
+	for _, gnb := range procReq.SiteInfo.GNodeBs {
+		if !isValidName(gnb.Name) {
+			err := fmt.Errorf("invalid gNB name `%s` in Network Slice %s. Name needs to match the following regular expression: %s", gnb.Name, sliceName, NAME_PATTERN)
+			logger.ConfigLog.Errorln(err.Error())
+			return err
+		}
+		castedGnbTac := fmt.Sprint(gnb.Tac)
+		if !isValidGnbTac(castedGnbTac) {
```
gab-arrobo, line 210: Alternatively, would not it be easier/faster to avoid the conversion (and avoid calling `isValidGnbTac`) and just check the following, which is basically what `isValidName` is doing "under the hood"
```suggestion
		if gnb.Tac < 1 && gnb.Tac > 16777215 {
```

dariofaccin: I would change `Tac` in `PutGnbRequest` to `int32`, howeverr this is a breaking change which changes the behaviour of the API. If you are ok with that, I'll proceed.

configapi/api_inventory.go
```diff
@@ -90,9 +89,9 @@ func PostGnb(c *gin.Context) {
 		c.JSON(http.StatusBadRequest, gin.H{"error": errorMessage})
 		return
 	}
-	if postGnbParams.Tac != "" {
+	if postGnbParams.Tac != 0 {
```
patriciareinoso, line 92: on this line we would be writing tac = 0 in the DB if the user explicitly sets the value to 0

patriciareinoso (changes requested):
please run 
```
swag init -g backend/webui_service/swagger_ui_service.go --outputTypes go
```
to update the  swagger ui docs

configapi/api_slice_mgmt.go
gab-arrobo: I think the API is mainly used by `Aether ROC/simapp` and `Canonical's NMS`, we need to validate that nothing breaks with these two because of the API change

dariofaccin: @gab-arrobo regarding Canonical NMS, we already have the PR to address this. Regarding `simapp`, it doesn't seem to use the `/api/config/v1/inventory/gnb` APIs, so this PR does not affect its behavior; it creates a NS but there the value for the TAC is already `1` (int) so the validation is positive.

configmodels/model_inventory.go
```diff
@@ -10,16 +10,16 @@ const (
 
 type Gnb struct {
 	Name string `json:"name"`
-	Tac  string `json:"tac,omitempty"`
+	Tac  *int32 `json:"tac,omitempty"`
```
gab-arrobo, line 13: Why does this `Tac` have to be a pointer and the one in `PutGnbRequest` is not?

configmodels/model_inventory.go
```diff
@@ -10,16 +10,16 @@ const (
...
 }
 
 type PostGnbRequest struct {
 	Name string `json:"name"`
-	Tac  string `json:"tac,omitempty"`
+	Tac  *int32 `json:"tac"`
```
gab-arrobo, line 18: Why does this `Tac` have to be a pointer and the one in `PutGnbRequest` is not?

gab-arrobo (commented):
Overall, it looks good to me. I have two minor comments/questions. I need to run a test using the ROC to make sure everything works as expected. Just in case, did you run a test using the ROC?

gab-arrobo (approved):
+1

configmodels/model_inventory.go
dariofaccin: In the POST method we now allow the creation of a gNB without specifying a TAC: this is required for the network configuration orchestration using Canonical charms (the PR has been merged days ago). The necessity of using a pointer arises from the fact that we need to distinguish between a non-set value (allowed) and an explicitly-set `0` value (not allowed): the only way to deal with this case in Go is to use pointers instead of direct values. I added a unit test for this edge case, which fails if using `int32`.

PR #332, opened 2025-03-18
fix: duplicate user account returns 409 conflict

patriciareinoso:
This PR modifies the Create User account operation to return a 409 Conflict error if the user account already exists instead of a 400 Bad request


gab-arrobo (approved):
+1

PR #341, opened 2025-04-02
chore: enable staticcheck linter in CI

patriciareinoso:
This PR enables `staticcheck` linter in the CI and makes the modifications needed in the code for it to pass:
- Fix variable declarations
- Replace if with switch


gab-arrobo (approved):
+1

PR #347, opened 2025-04-07
chore: improve name validation for gNBs, Network Slices, Device Groups

dariofaccin:
This PR adds the name length validation for gNBs, Network Slices and Device Groups.
The new regular expression (`^[a-zA-Z][a-zA-Z0-9-_]{1,255}$`) limits the length to 256.


configapi/validators.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	NAME_PATTERN = "^[a-zA-Z][a-zA-Z0-9-_]+$"
+	NAME_PATTERN = "^[a-zA-Z][a-zA-Z0-9-_]{1,255}$"
```
gab-arrobo, line 12: Out of curiosity... is the intention to force the name to be between 2 and 256 characters?

dariofaccin: Currently the name must be at least 2 characters (the first character must be alphabetic `a-zA-Z`; the operator `+` matches between one and unlimited times the following: `a-zA-Z0-9-_`), with no maximum length limit. This PR enforces a maximum length of 256.

gab-arrobo (approved):

PR #350, opened 2025-04-18
Fix issue with missing inherit secrets

gab-arrobo:


thakurajayL:
What does inherit secret do?

gab-arrobo:
> What does inherit secret do?

For this specific task, we publish a release/tag in GitHub and to be able to do so, we need a GitHub token that is configured as secret. Since now we are using "centralized" actions from `omec-project/.github`, we need that the action inherits the secret such that it has the permissions to create the tag

thakurajayL (approved):

PR #390, opened 2025-07-18
feat: Implements Policy Control and IMSI QoS endpoints

Gmerold:
This PR implements Policy Control and IMSI QoS endpoints according to the new NF configuration system design.
Both endpoints are implemented in the same PR, because they are somewhat related. 

In both cases it is assumed that the configuration parameters of Network Slices and Device Groups have been validated upon creation (POST) or update (PUT) of a Network Slice/Device Group. As indicated in the inline comments, the validation is not implemented yet.


thakurajayL:
@Gmerold could you please share more about what is intention of this PR. What end goal it achieves? Could you share the pointer to the design which you referred in the PR description? 

Thanks for your contribution.

gab-arrobo (changes requested):
Please check whether this properly works with an E2E test because the E2E test fails in my local setup. The webconsole crashes:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x14 pc=0x1011734]

goroutine 1 gp=0xc000002380 m=6 mp=0xc000395008 [running]:
panic({0x1236160?, 0x1ec88e0?})
        /usr/local/go/src/runtime/panic.go:811 +0x168 fp=0xc0006cd010 sp=0xc0006ccf60 pc=0x478f48
runtime.panicmem(...)
        /usr/local/go/src/runtime/panic.go:262
runtime.sigpanic()
        /usr/local/go/src/runtime/signal_unix.go:925 +0x359 fp=0xc0006cd070 sp=0xc0006cd010 pc=0x47b7b9
github.com/omec-project/webconsole/backend/nfconfig.buildPccQos({{0xc00049f0d0, 0x9}, 0xfa, {0xc00049f0a8, 0x6}, {0xc00049f0c0, 0x9}, 0x0, 0x0, 0x0, ...})
        /go/src/webconsole/backend/nfconfig/config.go:440 +0x74 fp=0xc0006cd140 sp=0xc0006cd070 pc=0x1011734
github.com/omec-project/webconsole/backend/nfconfig.buildSlicePccRules({{0xc00049f126, 0x7}, {{0x161bc68, 0x1}, {0xc00049f120, 0x6}}, {0xc000384ff0, 0x1, 0x1}, {{0xc00049f0f0, ...}, ...}, ...})
        /go/src/webconsole/backend/nfconfig/config.go:371 +0x178 fp=0xc0006cd3e8 sp=0xc0006cd140 pc=0x1010d38
github.com/omec-project/webconsole/backend/nfconfig.buildPolicyControlConfig({{0xc00049f126, 0x7}, {{0x161bc68, 0x1}, {0xc00049f120, 0x6}}, {0xc000384ff0, 0x1, 0x1}, {{0xc00049f0f0, ...}, ...}, ...})
        /go/src/webconsole/backend/nfconfig/config.go:354 +0x198 fp=0xc0006cd548 sp=0xc0006cd3e8 pc=0x1010a98
github.com/omec-project/webconsole/backend/nfconfig.(*inMemoryConfig).syncPolicyControl(0xc00052c590, {0xc0003044d0?, 0x1?, 0xc00014c0a8?})
        /go/src/webconsole/backend/nfconfig/config.go:336 +0xdd fp=0xc0006cd718 sp=0xc0006cd548 pc=0x101069d
github.com/omec-project/webconsole/backend/nfconfig.(*NFConfigServer).syncInMemoryConfig(0xc00052c580)
        /go/src/webconsole/backend/nfconfig/service.go:198 +0x70b fp=0xc0006cd998 sp=0xc0006cd718 pc=0x101386b
github.com/omec-project/webconsole/backend/nfconfig.NewNFConfigServer(0xc000130378)
        /go/src/webconsole/backend/nfconfig/service.go:67 +0x1c8 fp=0xc0006cda28 sp=0xc0006cd998 pc=0x1012528
main.startApplication(0xc000130378)
        /go/src/webconsole/server.go:76 +0xd3 fp=0xc0006cdab8 sp=0xc0006cda28 pc=0x10cc733
main.action({0xc000158008?, 0x15fa318?}, 0xc000503620?)
        /go/src/webconsole/server.go:64 +0xd9 fp=0xc0006cdb08 sp=0xc0006cdab8 pc=0x10cc5d9
```

backend/nfconfig/config.go
```diff
@@ -6,11 +6,15 @@
 package nfconfig
 
 import (
+	"fmt"
 	"slices"
 	"sort"
 	"strconv"
+	"strings"
 
+	protos "github.com/omec-project/config5g/proto/sdcoreConfig"
```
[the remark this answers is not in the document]
patriciareinoso: i agree, we should not be using the `protos` package anymore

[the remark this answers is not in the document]
Gmerold: Yes, I accidentally moved it with the rest of the flow description generation logic.
It's removed.

Gmerold:
> > In that context, I think the E2E tests will have to be updated to match this assumption. Is that ok with you?
> 
> What would be your plan to update the E2E tests? What would need to be modified/updated?

I'm not super familiar with your E2E setup, but after a brief check it seems to me that the changes should land [here](https://github.com/opennetworkinglab/aether-5gc/blob/4180bba19c1db41ef3a88bc3b705f25dfca2c9d9/roles/core/templates/radio-5g-values.yaml#L196). What we need is a TrafficClass (Qci and Arp specifically) to be filled. 
In the current implementation of the `webconsole`, if these values are not there, defaults are applied by the `fillSlice` function. The idea in the long run is to do this upon Slice creation/update.

backend/nfconfig/handlers.go
```diff
@@ -36,3 +38,16 @@ func (n *NFConfigServer) GetSessionManagementConfig(c *gin.Context) {
...
 }
+
+func (n *NFConfigServer) GetImsiQosConfig(c *gin.Context) {
+	dnn := c.Param("dnn")
+	imsi := c.Param("imsi")
+	logger.NfConfigLog.Debugf("Handling GET request for QoS config for IMSI %s", imsi)
+	var imsiQos []nfConfigApi.ImsiQos
+	for _, imsiQosConfig := range n.inMemoryConfig.imsiQos {
+		if imsiQosConfig.dnn == dnn && slices.Contains(imsiQosConfig.imsis, imsi) {
+			imsiQos = imsiQosConfig.qos
```
patriciareinoso, line 49: we are returning only the last element?

Gmerold: There should be only one matching element anyway, but the code is refactored now to break after a hit.

patriciareinoso (approved):

gab-arrobo:
> > > In that context, I think the E2E tests will have to be updated to match this assumption. Is that ok with you?
> > 
> > 
> > What would be your plan to update the E2E tests? What would need to be modified/updated?
> 
> I'm not super familiar with your E2E setup, but after a brief check it seems to me that the changes should land [here](https://github.com/opennetworkinglab/aether-5gc/blob/4180bba19c1db41ef3a88bc3b705f25dfca2c9d9/roles/core/templates/radio-5g-values.yaml#L196). What we need is a TrafficClass (Qci and Arp specifically) to be filled. In the current implementation of the `webconsole`, if these values are not there, defaults are applied by the `fillSlice` function. The idea in the long run is to do this upon Slice creation/update.

```go
func buildPccQos(ruleConfig configmodels.SliceApplicationFilteringRules) nfConfigApi.PccQos {
	pccQos := nfConfigApi.NewPccQos(
		ruleConfig.TrafficClass.Qci,
		configapi.ConvertToString(uint64(ruleConfig.AppMbrUplink)),
		configapi.ConvertToString(uint64(ruleConfig.AppMbrDownlink)),
		*nfConfigApi.NewArp(
			ruleConfig.TrafficClass.Arp,
			nfConfigApi.PREEMPTCAP_MAY_PREEMPT,
			nfConfigApi.PREEMPTVULN_PREEMPTABLE,
		),
	)
	return *pccQos
}
```

Based on your `buildPccQos`, you would require `traffic-class` to be different than nil. So, by making the change below in OnRamp, the E2E tests work as expected. However, `AppMbrUplink` and `AppMbrDownlink` would be `0`. What would be the side-effect of this? Do these two values have to be defined and be set as different than zero (and also define `BitrateUnit`)? I have not look into the details on how this is actually being used in the NFs (PCF) or how are you planning to use/set this in the PCF? In your setup, what is required and what is the default value?
```diff
diff --git a/roles/core/templates/sdcore-5g-values.yaml b/roles/core/templates/sdcore-5g-values.yaml
index 9a03d7f..2fe6818 100644
--- a/roles/core/templates/sdcore-5g-values.yaml
+++ b/roles/core/templates/sdcore-5g-values.yaml
@@ -306,6 +306,9 @@ omec-sub-provision:
                 priority: 250
                 action: "permit"
                 endpoint: "0.0.0.0/0"
+                traffic-class:
+                  qci: 9
+                  arp: 6
               site-info:
                 # Provide gNBs and UPF details and also PLMN for the site
                 gNodeBs:
```

@thakurajayL, @ghislainbourgeois, what are your thoughts about this?

backend/nfconfig/config.go
```diff
@@ -296,7 +334,146 @@ func extractGnbNames(slice configmodels.Slice) []string {
...
+	} else {
+		return fmt.Sprintf("permit out %s from %s to assigned %s-%s", protocol, endpoint, strconv.FormatInt(int64(startPort), 10), strconv.FormatInt(int64(endPort), 10))
+	}
+}
+
+func buildPccQos(ruleConfig configmodels.SliceApplicationFilteringRules) nfConfigApi.PccQos {
+	pccQos := nfConfigApi.NewPccQos(
+		ruleConfig.TrafficClass.Qci,
+		configapi.ConvertToString(uint64(ruleConfig.AppMbrUplink)),
+		configapi.ConvertToString(uint64(ruleConfig.AppMbrDownlink)),
```
gab-arrobo, line 442: I think you need to take into account `BitrateUnit`. That is, these should be like:
```suggestion
		strconv.FormatInt(int64(ruleConfig.AppMbrUplink), 10) + ruleConfig.BitrateUnit,
		strconv.FormatInt(int64(ruleConfig.AppMbrDownlink), 10) + ruleConfig.BitrateUnit,
```

gab-arrobo (changes requested):
It looks like it is assumed that the `bit rates` are given in `bps` but actually there is a field (`BitrateUnit`), where users can indicate that. Also, I see there is no validation of these fields. So, I assume this will be done later as you indicate in the `buildSlicePccRules` function, correct?
```go
func buildSlicePccRules(slice configmodels.Slice) []nfConfigApi.PccRule {
	/* Implementation assumes that the validation of a Network Slice configuration is done upon group creation/modification.
	At the time of implementing this, validation is not done, but planned.

	TODO: Remove this comment once Device Group validation is implemented.
	*/
...
...
```

backend/nfconfig/config.go
Gmerold: Hello @gab-arrobo,
The values in the DB are actually always in `bps`. POST helpers convert values given by the user to `bps`. It's true that the `BitrateUnit` can indicate different units, but this is actually a bug - bitrates are converted to bps before saving to the DB, but the unit given by the user is preserved. As a result, if you POST a DevGroup with Uplink bitrate set to 200 Mbps, GET will return 200000000 Mbps. 
This bug should be addressed separately. Until it's fixed, I think it's safer to assume the value is in `bps` and ignore the `BitrateUnit` field.

gab-arrobo (approved):
+1

PR #406, opened 2025-08-13
fix: upf port can be a int or float64

patriciareinoso:
# Description

Given that the UPF field in the network slices is defined as 
```
	Upf map[string]interface{} `json:"upf,omitempty"`
```
and can be initialized as:
```
 map[string]interface{}{
			"upf-name": "some hostname",
			"upf-port": port,
		}
```

It is not clear what the type for the `upf-port` is. 

This PR extends the possibility of the port to also be a float64 (default type when unmarshaling), int.


gatici (approved):

gab-arrobo (approved):
+1

